PR #4, opened 2015-04-09
Use Control.Monad.Ref from the ref-tf package

bennofs:
The ref-tf package already contains classes for this purpose and also
makes instances for transformer types. It also introduces an extra type
classes for atomic updates.

This removes some of the extra functions that the Control.Monad.Ref
module provided (memoM, cacheM, cacheMWithTry, replaceRef). Should those be
re-added in a different module (which?)?

I'm not sure what to do about the SPECIALIZE pragma for the ReaderT instance, since ref-tf 
doesn't contain that. Do you have evidence that it is needed for good performance?


ryantrinkle:
Thanks!  I've pulled this onto develop as 94ba8c8a575aed16856688d5856be82089e3d33c; it'll be in the next release.

ryantrinkle:
To answer some of your questions:

I'm not sure whether the SPECIALIZE is important.  I'll look into the performance, and perhaps we can get them to accept it upstream if it makes a big difference.

memoM, cacheM, etc., don't appear to be used in the current codebase, so I think it's fine to get rid of them.

PR #9, opened 2015-04-29
Docs for Reflex.Class

cocreature:
While reading the source I thought I could just as well add some docs so here they are. I might have missunderstood some parts so better read through it :) Because of lazyness I also put the removal of 'sequenceThese' in the same commit (tell me if I should remove this). This is exactly 'bisequenceA' from 'Bitraversable' which 'These' is an instance of.


ryantrinkle:
sequenceThese should be deprecated rather than removed, but the rest looks good.  Thanks for putting all this effort into the docs!

cocreature:
Oh rebasing on develop and force pushing doesn't seem to have been such a good idea as I accidentally made the pr against master. Do you manage this when merging or should I make another pr against develop? I only changed the deprecation stuff.

ryantrinkle:
I'll handle it in the merge.

ryantrinkle:
Is there a particular reason to suggest bisequenceA instead of bisequence?  It looks like bisequence has a closer type signature to the original.

cocreature:
The reason is that bisequenceA is more general and at least with 7.10 always works where bisequence works due to AMP. And even before 7.10 there is a warning if your Monad is not an Applicative so this seems pretty save to me. If you disagree I can of course change it/you just change it when merging.

ryantrinkle:
Merged; thanks!

PR #10, opened 2015-04-29
Remove two warnings

cocreature:
Should be pretty obvious. As you are using `-Wall` it seems like a good idea to get rid of the warnings where possible.


ryantrinkle:
Merged to develop; thanks!

PR #14, opened 2015-05-16
Bump nix package to 0.2

stighenriksen:


ryantrinkle:
Oh, good catch; thanks!

PR #15, opened 2015-06-15
Add nubDynBy function

mightybyte:
This adds a pretty straightforward and useful nubDynBy function.  I wasn't able to include it in my project because it uses the Dynamic constructor that is not exported.

This branched off another branch I created with a small fix for a GHC 7.8 build error.


ryantrinkle:
This has been added [here](https://github.com/reflex-frp/reflex/blob/functor-dynamic/src/Reflex/Dynamic.hs#L96), and will be merged to develop soon.

PR #20, opened 2015-07-13
Fixes for GHC 7.8

bennofs:
This PR fixes some things for GHC 7.8:

1. GHC 7.8 comes with an older transformers, which doesn't have ExceptT yet. This PR 
adds a dependency on transformers-compat, which provides ExceptT, to solve this.
2. GHC 7.8 didn't implement AMP yet, so we need to provide explicit Applicative 
constraints in some places
3. To silence warnings about unneccessary imports of Control.Applicative, Prelude is 
imported last


ryantrinkle:
Thanks!

PR #23, opened 2015-07-31
Add Quickref.md

dbbnrl:


Quickref.md
```diff
@@ -0,0 +1,170 @@
+# Reflex Quick(ish) Reference
+
+## Typeclasses
+
+Many Reflex functions operate in monadic context `X a`, where the monadic type X supports various additional typeclasses such as MonadWidget, MonadHold, or MonadSample in addition to Monad itself.  For Reflex-Dom users, the type X will generally be `Widget Spider (Gui Spider (WithWebView SpiderHost) (HostFrame Spider))`, from the Reflex-Dom main entry point:
+
+```haskell
+mainWidget :: Widget Spider (Gui Spider (WithWebView SpiderHost) (HostFrame Spider)) () -> IO ()
```
mightybyte: This type signature is a bit scary and you don't actually have to use it.  In fact, the string "Spider" doesn't appear a single time in my 7k line reflex application.  I just use the simpler:

    fooWidget :: MonadWidget t m => m ()

So perhaps it would be better to change it to use this more generic and concise signature?

dbbnrl: Agreed.  The concrete type is listed at the bottom of the reflex-dom quickref, but there's no need to repeat it here.  I'll change it.

ryantrinkle: I would recommend listing the type as:
mainWidget :: (forall t m. MonadWidget t m => m ()) -> IO ()

I will probably change mainWidget to **actually** have that type, soon.

mightybyte: Oh, but it probably shouldn't be MonadWidget because that type class is defined in reflex-dom.  But I think there's another set of type class constraints that works here that would be better than the Widget Spider stuff.

dbbnrl: I think the point is that there's no need to refer to the type of mainWidget *at all* in the Reflex quickref.

Quickref.md
```diff
@@ -0,0 +1,168 @@
...
+
+-- Flatten Event-of-Event to Event that fires when both wrapper AND new Event fire.
+[ ]   coincidence       ::                     Event (Event a)  ->    Event a
+
+-- Flatten Dynamic-of-Dynamic to Dynamic.  New Dynamic is used immediately.
+-- Output updated whenever inner OR outer Dynamic updates. 
+[ ]   joinDyn           ::          Dynamic        (Dynamic a)  ->  Dynamic a
+[ ]   joinDynThroughMap :: Ord k => Dynamic (Map k (Dynamic a)) ->  Dynamic (Map k a)
+
+-- Flatten Event-of-Behavior to monadic Behavior
```
ryantrinkle: Should probably say something like "Create a Behavior whose value initially tracks the first argument.  When the Event fires, that Behavior is replaced, and the output tracks the new Behavior until it the Event fires again."

Quickref.md
```diff
@@ -0,0 +1,168 @@
...
+[H]   combineDyn :: (a -> b -> c) -> Dynamic a -> Dynamic b -> m (Dynamic c)
+
+-- Efficient one-to-many fanout
+[ ]   demux      :: Ord k => Dynamic k -> Demux k
+[H]   getDemuxed ::  Eq k =>              Demux k -> k -> m (Dynamic Bool)
+
+-- Dynamic to Dynamic, removing updates w/o value change
+[ ]   nubDyn :: Eq a => Dynamic a -> Dynamic a
+
+-- Dynamic to identical Dynamic with debug trace.
```
ryantrinkle: Note that with any of the trace functions, the value will only be traced if the output is actually used; for traceDyn and traceEvent, that means it only traces when someone's actually listening to the output event.  For traceBehavior, it means that it'll only trace when the Behavior's value is actually sampled.

Quickref.md
```diff
@@ -0,0 +1,168 @@
...
+```
+
+## Flattening functions
+
+These functions flatten nested types such as Event-of-Event, Behavior-of-Event, Event-of-Behavior etc, removing the outer wrapper.
+
+For Events, the returned Event fires whenever the latest Event supplied by the wrapper fires.  There are differences in how the functions handle *coincidences* -- situations where the old or new Event fires at the same instant that we switch from old to new.  In some cases, the output Event tracks the old Event at the instant of switchover, while in other cases it tracks the new Event.
+
+```haskell
+-- Flatten Event-of-Event to monadic Event.  New Event is used immediately.
```
ryantrinkle: I think we can drop the "to monadic Event" part - perhaps it's better to think about it as flattening starting at a particular point in time.

PR #24, opened 2015-08-01
instanceSigs

jeffreyrosenbluth:
Is it necessary to use `memo` in the pure implementation?


ryantrinkle:
We need `memo` so that the tests ensuring Pure and Spider produce the same results can be run relatively efficiently.  Perhaps we should add some comment telling readers they can ignore it?  `memo` is semantically equivalent to `id`.

jeffreyrosenbluth:
Hmmm, I'm surprised that the tests are complex enough to create performance issues.

PR #25, opened 2015-08-31
Add instances for strict StateT

Saulzar:
Really only needed the MonadReflexCreateTrigger which keeps hanging around as an orphan instance in my code, but the others were there so added them too.


ryantrinkle:
Thanks!

PR #35, opened 2015-11-30
Update to use dependent-map-0.2.* and dependent-sum-0.3.*

cgibbard:


ryantrinkle:
Just closing this out because the work was merged separately from this PR.  Thanks @cgibbard!

PR #36, opened 2015-12-03
add combineDyn3 and combineDyn4

mulderr:
I'm putting this here for consideration.

I find that sometimes I need a `Dynamic` that is derived from 3 or more other `Dynamic`s.

`mkDyn` is very flexible itself so one may want to just use it directly and it seems having `combineDynN` explicitly defined adds little value. Also stopping at 4 is totally arbitrary - it's just that currently I don't have a use case for more.

Cons:
* cannot add to `Reflex.Dynamic` - would cause circular imports between `Reflex.Dynamic` and `Reflex.Dynamic.TH`
* will add to library compilation time with ghcjs

Pros:
* nicer syntax - don't have to explicitly use TH
* if the pattern turns out to be common enough it can save lots of compilation time by shifting the TH burden to the library - using `mkDyn` adds a good couple seconds on each compile for my test project

It depends on how often do people currently need to chain `combineDyn`s. Eg. `combineDyn (,) da db >>= combineDyn f dc` could be replaced by `combineDyn3 g da db dc` without affecting compilation time.

So main question: would anyone else find it useful?

### DIFF
--- a/reflex.cabal
+++ b/reflex.cabal
@@ -39,6 +39,7 @@ library
     Reflex.Spider.Internal,
     Reflex.Class,
     Reflex.Dynamic,
+    Reflex.Dynamic.Combinators,
     Reflex.Dynamic.TH,
     Reflex.Host.Class,
     Data.Functor.Misc
--- a/src/Reflex.hs
+++ b/src/Reflex.hs
@@ -1,10 +1,12 @@
 module Reflex ( module Reflex.Class
               , module Reflex.Dynamic
+              , module Reflex.Dynamic.Combinators
               , module Reflex.Dynamic.TH
               , module Reflex.Spider
               ) where
 
 import Reflex.Class
 import Reflex.Dynamic
+import Reflex.Dynamic.Combinators
 import Reflex.Dynamic.TH
 import Reflex.Spider
--- a/src/Reflex/Dynamic/Combinators.hs
+++ b/src/Reflex/Dynamic/Combinators.hs
@@ -1,0 +1,29 @@
+{-# LANGUAGE QuasiQuotes, GADTs #-}
+module Reflex.Dynamic.Combinators ( combineDyn3
+                                  , combineDyn4
+                                  ) where
+
+import Reflex.Class
+import Reflex.Dynamic
+import Reflex.Dynamic.TH
+
+-- | combineDyn for three 'Dynamic's.
+combineDyn3 :: (Reflex t, MonadHold t m)
+  => (a -> b -> c -> d)
+  -> Dynamic t a
+  -> Dynamic t b
+  -> Dynamic t c
+  -> m (Dynamic t d)
+combineDyn3 f a b c = [mkDyn|f $a $b $c|]
+
+-- | combineDyn for four 'Dynamic's.
+combineDyn4 :: (Reflex t, MonadHold t m)
+  => (a -> b -> c -> d -> e)
+  -> Dynamic t a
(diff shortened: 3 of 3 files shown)


mightybyte:
The standard approach for doing this is `mkDyn` in [Reflex.Dynamic.TH](http://hackage.haskell.org/package/reflex-0.3.2/docs/Reflex-Dynamic-TH.html).  Using mkDyn to do your combineDyn3 example would look like this:

    -- At the top of your file
    {-# LANGUAGE QuasiQuotes #-}

    newDyn <- [mkDyn| g da db dc |]

mulderr:
Hmm... right, so no point to do it really except maybe the compilation time thingy which I admit is weak.

cgibbard:
Compilation time is pretty important to me. Waiting for our large project to recompile in cases where I'd made some change to a foundational module has been a major leak in productivity for me. mkDyn is clearly nicer to use, and we'll have to see what happens when we switch to using the new ghcjs, but I do think the difference in compile time for modules that have several occurrences of mkDyn in them is a serious argument.

mulderr:
@cgibbard can't really advocate it but as a workaround you could try to create a module with mkDyn wrappers inside your project. It would get compiled once and you would not get hit by TH stuff in other modules every time they change. It may be a bit of work and you may just end up going back in a while but if it's already an annoyance...

After some thought I don't think there is reason enough to expose this as public API in reflex now when, as I understand it, there is a decent probability it will just have to be deprecated when ghcjs improvements become available.

mightybyte:
I was just working on some code and a perfect situation for testing this presented itself.  The results are alarming enough that I thought it would be good to record it for posterity and this seemed as good a place as any.  I had a pretty straightforward 270 line module we'll call Foo.  I changed a function that took two Dynamic params and returned another Dynamic to instead take three Dynamic params.  I already had the dynamic available, so literally the only change was to add the parameter to the non-dynamic function, change a combineDyn to a mkDyn, and add the QuasiQuotes pragma.  I timed the build before and after the change by hand using a stopwatch.  Here are the results:

combineDyn: 2.3 seconds
mkDyn: 1 minute 39 seconds!!!!

My timing methodology was to start the stopwatch by hand as immediately after seeing this line appear:

    [67 of 76] Compiling Foo ...

I stopped the stopwatch as soon as I saw the `[68 of 76]` line signaling that it had finished compiling the Foo module and had moved on to the next one.  There's certainly some error in my times, but it's almost certainly less than plus/minus half a second.

I'm not posting this in an attempt to reopen this issue.  I don't know whether this is a GHCJS issue or an issue with the implementation of mkDyn.  I just thought the severity of this problem should be mentioned somewhere.

cgibbard:
Yes. This agrees very well with my experience, and it's basically why I made the comment that I did. Most of the time I'd be waiting on a compile on one of our larger projects was spent linking Template Haskell. For the whole project, it could easily get up to 10 or even 15 minutes sometimes depending on which part of the frontend I was working on, or if I was working on code common to the frontend and backend, to the point that I basically had to have compiles going in the background all the time while I looked for anything that was likely to result in a type error or continued on with the next piece of work, hoping that the compile for the previous thing didn't fail. Using ghcid improved the situation by a lot (it can report errors in a much shorter timespan than it takes to do all the TH linking), but it's still not as nice as just having the real compile be fast.

The compile times are basically my largest (and almost sole) source of frustration with using reflex-dom in that setting at this point, so I really hope that the new ghcjs does something dramatic about them.

PR #37, opened 2015-12-04
Change the variable name used when generating mkDyn expressions to something far less likely to capture.

cgibbard:
I don't know if Template Haskell has a better solution to this problem, but this at least keeps people from
stumbling over dyn.


bennofs:
I thought that newName in TH already generates a non-capturing name automatically? Is that wrong?

cgibbard:
According to the documentation it creates a name which isn't capturable, but which may capture other names. The name in question here was being identified with dyn from reflex-dom.

PR #39, opened 2016-01-06
Functor instance for Dynamic

snoyberg:
As I'm very new to reflex, this may be a bad idea for reasons I don't understand. However, in my usage so far, such an instance would have been convenient (as an alternative to `y <- mapDyn f x`).

### DIFF
--- a/src/Reflex/Dynamic.hs
+++ b/src/Reflex/Dynamic.hs
@@ -93,6 +93,11 @@ instance HBuild' (a ': l) r
 -- | A container for a value that can change over time and allows notifications on changes.
 -- Basically a combination of a 'Behavior' and an 'Event', with a rule that the Behavior will
 -- change if and only if the Event fires.
+--
+-- Although @Dynamic@ logically has a 'Functor' instance, currently it can\'t
+-- be implemented without potentially requiring the mapped function to be
+-- evaluated twice. Instead, you can use 'mapDyn', which runs in a 'MonadHold'
+-- instance, to achieve the same result.
 data Dynamic t a
   = Dynamic (Behavior t a) (Event t a)
 


bennofs:
The problem with this instance is that it will evaluate `f` twice whenever the input dynamic changes: once to compute the new `Event` value, and once to compute the new `Behavior` value. With `mapDyn`, there is only a single computation, the result of both is shared. This is also why `mapDyn` is monadic.

ryantrinkle:
@snoyberg It's great to get your feedback on this.

I completely agree with what you're saying - it would be way nicer to use `fmap` than `mapDyn`.  However, @bennofs basically covered it.  Currently, there's no way to implement `Dynamic` without potentially losing sharing.  This is something I plan to fix in the near future.

snoyberg:
That reasoning makes more sense. A few thoughts from me (and, again, I realize these are very beginner-level questions):

* In many of the cases where I would have wanted to use a `Functor` instance, I would not have expected the lost sharing to be an issue. For example, if I use `dynText`, I would have thought that the initial rendering would use the `Behavior` and all future updates would use the `Event`, ignoring the `Behavior`, meaning that no value is ever used from both the `Behavior` and `Event`.
* Having this kind of information in the docs for `Dynamic` may be useful for people like me (used to Haskell typeclasses, and surprised when they aren't being used). Once we have clarity here, I'd be happy to update this PR with such a documentation change.

ryantrinkle:
I think you're right that, usually, the lack of sharing would not be an issue.  `dynText` certainly would not be impacted.  However, I've made an effort to ensure that reflex doesn't have any weird corner cases where the performance gets really bad, and I think this is a situation where it could.  I definitely wouldn't want someone to spend a bunch of time profiling their code, only to find that something like this is at its core.

Adding documentation makes sense to me.  What do you think of adding something to `Dynamic`'s docs saying something like "Although Dynamic logically has a Functor instance, currently it can't be implemented without potentially requiring the mapped function to be evaluated twice.  Instead, you can use mapDyn, which runs in a MonadHold instance, to achieve the same result."?

snoyberg:
That sounds perfect. As a reader, that says to me:

* Aha! My understanding of this concept is correct, I'm not crazy.
* There's a deeper reason for the lack of this instance, which is rightfully beyond my ability to predict, but the docs have helpfully told me how to achieve the same thing in a different way.

Big :+1: :)

chrisdone:
:+1:

snoyberg:
I've updated the PR to include your suggested documentation.

ryantrinkle:
Awesome!  Thanks, @snoyberg!

Issue #19, opened 2015-07-09
FunctorMaybe


andrewthad:
I feel like `FunctorMaybe` isn't the most useful typeclass. It's got one instance, and that let's us call things like `ffilter` on `Event`s. It seems like it would be better to just have functions `filterEvent` and `filterMapEvent` instead of `ffilter` and `fforMaybe`. They would be more monomorphic and more understandable.

mightybyte:
I'm not sure of how important the type class is, but I use fmapMaybe all the time.  Removing those functions would break a lot of code.  They also feel more fundamental to me than the filter variants.

ryantrinkle:
I would much prefer to reuse an existing typeclass with this functionality rather than creating my own - I agree that it's not terribly useful so long as it's living in this library and only used by Event.  However, I asked around and I wasn't able to find something that already existed.  I'd be happy to switch to using an external typeclass if there's one that provides this functionality.  Please feel free to reopen this issue if you have one in mind.

bennofs:
@ryantrinkle How about http://hackage.haskell.org/package/filtrable-0.1.0.2/docs/Data-Filtrable.html ?

andrewthad:
Yeah. That one seems like a pretty solid choice.

andrewthad:
Also, even better than the `filterable` package is [witherable](https://hackage.haskell.org/package/witherable-0.1.3.2/docs/Data-Witherable.html)

ryantrinkle:
Looks reasonable. @bennofs; any thoughts?

bennofs:
@andrewthad @ryantrinkle `witherable` won't work for us, since `Event` cannot be `Traversable` (which `witherable` requires as a superclass and the `wither` function can be used to implement `traverse` I believe). The reason for that is that there is no way to get infomation "out" of an Event (think of `traverse` with the `Const` applicative).

spl:
I also agree that the `FunctorMaybe` is a strange beast, being in `Reflex.Class` and having only one instance. But since it is useful and already there, I would prefer that it remain that way until something  decidedly better comes along.

As for `witherable`, I wonder if the `Traversable` superclass is really necessary or just a convenience. It's only used here in the class definition:

```haskell
class Traversable t => Witherable t where
  wither :: Applicative f => (a -> f (Maybe b)) -> t a -> f (t b)
  wither f = fmap catMaybes . traverse f
  [...]
```

[...]

PR #44, opened 2016-02-11
Use Data.Filtrable instead of FunctorMaybe

ali-abrar:
Based on the discussion in #19

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,5 +1,5 @@
 { mkDerivation, dependent-map, dependent-sum
-, mtl, ref-tf, semigroups, these, MemoTrie, exception-transformers
+, mtl, ref-tf, semigroups, these, MemoTrie, exception-transformers, filtrable
 , haskell-src-exts, haskell-src-meta
 }:
 mkDerivation {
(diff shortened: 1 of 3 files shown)


ali-abrar:
I did a little looking around to see if any other package would be better for what we need and didn't find anything, but suggestions are certainly welcome.

spl:
:+1:

3noch:
Are we aiming for American or English spellings? `Filtrable` is not the American spelling.

spl:
I didn't even realize “Filtrable” was a spelling in any dialect. Nonetheless, I think the issue that @ali-abrar is proposing a solution to is the singular use of `FunctorMaybe` instead of some more common or more standard type class. I wouldn't be against using `Filtrable` because it doesn't fit my expected spelling.

I think a better question is: what benefit would `reflex` get from using `filtrable` instead of `FunctorMaybe`? If it was a widely used package, then that would mean more people would be familiar with it and/or be able to use it in other places. Unfortunately, this is not the case according to [its reverse dependencies](http://packdeps.haskellers.com/reverse/filtrable). Still, I like the fact that somebody created a package that could potentially become more common.

3noch:
@spl Ah, I didn't even look at the diff so I didn't realize he was pulling in a dependency with that name! In that case, there's not much of a choice. Carry on!

ali-abrar:
I'm not completely satisfied with the use of this library either. I'm still looking around to see if there's something else we could use. `Data.Witherable` has been recommended by a few people, but the `Traversable` requirement it imposes is something we can't meet here (and, despite the recommendations, witherable's reverse dependencies reveal that it isn't really that much more popular than Filtrable).

ryantrinkle:
It would be great to replace FunctorMaybe with something more general - or at least more widely used - but it looks like Filtrable probably isn't that thing.  I'm going to close this, but hopefully another alternative will appear eventually.

PR #45, opened 2016-03-07
Add apDyn to enable applicative style programming on Dynamic

DiegoNolan:
Let me know your comments. Like if the haddock doc is too long or the type sig should be inline.


ryantrinkle:
Looks good; thanks!

PR #47, opened 2016-05-01
Add nubDynBy

waern:
This is just another little nudge to get this function in. I know there's already #15, but it hasn't been merged and I prefer if the function argument has type  a -> a -> Bool.

### DIFF
--- a/src/Reflex/Dynamic.hs
+++ b/src/Reflex/Dynamic.hs
@@ -5,6 +5,7 @@ module Reflex.Dynamic ( Dynamic -- Abstract so we can preserve the law that the 
                       , constDyn
                       , holdDyn
                       , nubDyn
+                      , nubDynBy
                       , count
                       , toggle
                       , switchPromptlyDyn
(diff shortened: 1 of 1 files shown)


ryantrinkle:
This function actually doesn't have quite the right semantics.  The problem is that the guarantee Dynamics give you is that the Behavior will change when, and only when, the Event fires.  With regular nubDyn, this is true, down to equality being correct.  However, if we allow users to supply a function, e.g. (\_ _ -> True), they can easily break this invariant.  The trouble is that there is no way currently to write nubDyn as a pure function while also preventing the Behavior from changing.  Within the next few weeks, there will be some new functionality that should make this possible - let's discuss this again then!

waern:
Oh, now I see why #15 was implemented the way it was. Happy to discuss this later.

ryantrinkle:
@waern `nubDyn` has been renamed `uniqDyn`, since it doesn't quite match the functionality of `Data.List.nub`, and `uniqDynBy` has also been added [here](https://github.com/reflex-frp/reflex/blob/functor-dynamic/src/Reflex/Dynamic.hs#L96).  Thanks for suggesting this!

PR #54, opened 2016-07-04
Fix base upper bound

nitrix:
Please bump the `base` upper bound.

As far as I can tell, reflex works fine on 4.9 and it's an inconvenience for some projects with many dependencies (like mine) where the only version of `base` they could agree on isn't supported by reflex.


nitrix:
From master, wrong branch.

PR #55, opened 2016-07-04
Fix template-haskell upper bound

nitrix:
Please bump the `template-haskell` upper bound.

As far as I can tell, reflex works fine on 2.12 and it's an inconvenience for some projects with many dependencies (like mine) where the only version of TH they could agree on isn't supported by reflex.


nitrix:
From master, wrong branch.

PR #58, opened 2016-07-18
More instances for `Behavior` and `Event`

esoeylemez:
This adds the following instances:

  * `Behavior`:
      * `Apply`
      * `Bind`
      * `Fractional`
      * `IsString`
      * `Num`
  * `Event`:
      * `Alt`
      * `Apply`
      * `Bind`
      * `Extend`
      * `Plus`

### DIFF
--- a/src/Reflex/Class.hs
+++ b/src/Reflex/Class.hs
@@ -1,3 +1,4 @@
+
 {-# LANGUAGE CPP #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE FlexibleInstances #-}
@@ -12,6 +13,7 @@
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE TypeOperators #-}
 {-# LANGUAGE UndecidableInstances #-}
+
 module Reflex.Class where
 
 import Control.Applicative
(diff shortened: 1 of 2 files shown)


src/Reflex/Class.hs
```diff
@@ -271,12 +301,38 @@ fforMaybe = flip fmapMaybe
...
+  evx >>- f = coincidence (f <$> evx)
+  join = coincidence
+
+-- | Event of single-occurrence events: @duplicated ev@ is an event that
+-- occurs at the same times as @ev@ carrying events that occur only once
+-- at that particular point in time.
+instance Reflex t => Extend (Event t) where
+  duplicated ev =
+      pushAlways (\_ -> switch <$> hold ev (never <$ ev))
+                 ev
```
ryantrinkle: This seems like a reasonable instance, but is there a reason that `duplicated e = e <$ e` wouldn't also be legitimate?  It would be good to document that if so.

esoeylemez:
Ryan, that instance doesn't seem to satisfy the `Extend` laws:

    Assumption: duplicated ev = ev <$ ev

    duplicated (duplicated ev)
        = duplicated (ev <$ ev)
        = (ev <$ ev) <$ (ev <$ ev)

    fmap duplicated (duplicated ev)
        = fmap duplicated (ev <$ ev)
        = fmap (duplicated . const ev) ev
        = fmap (const (duplicated ev)) ev
        = fmap (const (ev <$ ev)) ev
        = (ev <$ ev) <$ ev

src/Reflex/Class.hs
esoeylemez: Ryan, that instance doesn't seem to satisfy the `Extend` laws:

    Assumption: duplicated ev = ev <$ ev

    duplicated (duplicated ev)
        = duplicated (ev <$ ev)
        = (ev <$ ev) <$ (ev <$ ev)

    fmap duplicated (duplicated ev)
        = fmap duplicated (ev <$ ev)
        = fmap (duplicated . const ev) ev
        = fmap (const (duplicated ev)) ev
        = fmap (const (ev <$ ev)) ev
        = (ev <$ ev) <$ ev

ryantrinkle: Hmm, but isn't `x <$ (y <$ z) == x <$ z`?  Then `(ev <$ ev) <$ (ev <$ ev) == (ev <$ ev) <$ ev`.

esoeylemez: Oh, of course.

Well, I'm not sure which instance is the more useful one. I propose that we leave out the `Extend` instance entirely for now, until practical use cases for either variant pop up.

ryantrinkle: Sounds good.

esoeylemez: Done.

ryantrinkle:
Thanks!  This is really great work.

PR #60, opened 2016-08-10
Add Either and These event-splitting functions.

HectorAE:
These additions follow a common pattern for splitting a single event in two based on the inner type of the value. I believe they will reduce a lot of future boilerplate and set a standard practice for splitting events that can contain multiple types.


3noch:
I've written this myself locally before so 👍

ryantrinkle:
Thanks, Hector!

PR #61, opened 2016-08-10
Fix error with 'fanThese'.

HectorAE:
justThis and justThat produce Nothing in the 'These' case, so neither event produced by fanThese would fire when the value of the event is a 'These', exactly the opposite of what I intended.


ryantrinkle:
Thanks!

PR #66, opened 2016-09-28
Adds the `AttachApply` typeclass.

dalaing:
This is useful for combining several Behaviors or Dynamics
together and sampling them with an Event.

There is an example in the Haddocks for the typeclass that
might help explain further / motivate the addition.

### DIFF
--- a/src/Reflex/Class.hs
+++ b/src/Reflex/Class.hs
@@ -80,6 +80,8 @@ module Reflex.Class
   , mapAccumMaybe_
   , mapAccumMaybeM_
   , zipListWithEvent
+  , (<@>)
+  , (<@)
   , headE
   , tailE
   , headTailE
@@ -915,6 +917,76 @@ zipListWithEvent f l e = do
         h:t -> (Just t, Just $ f h b)
         _ -> (Nothing, Nothing) --TODO: Unsubscribe the event?
   mapAccumMaybe_ f' l e
+
+infixl 4 <@>
+
+-- | This is used to sample the value of a 'Behavior' using an 'Event'.
+--
+-- The '<@>' operator is intended to be used in conjunction with 
+-- the 'Applicative' instance for 'Behavior'.
+--
+-- This is useful when we want to combine the values of one 'Event' and 
+-- the value of several 'Behavior's at the time the 'Event' is firing.
+--
+-- If we have:
+--
+-- > f  :: a -> b -> c -> d
+-- > b1 :: Behavior t a
+-- > b2 :: Behavior t b
+-- > e  :: Event t c
+--
+-- then we can do:
+--
+-- > f <$> b1 <*> b2 <@> e :: Event t d
+--
+-- in order to apply the function 'f' to the relevant values.
+--
+-- The alternative would be something like:
+--
+-- > attachWith (\(x1, x2) y -> f x1 x2 y) ((,) <$> b1 <*> b2) e :: Event t d
+--
+-- or a variation involing a custom data type to hold the combination of 
+-- 'Behavior's even when that combination might only ever be used by 'f'.
+--
+-- A more suggestive example might be:
+--
+-- > handleMouse <$> bPlayerState <*> bMousePosition <*> eMouseClick :: Event t (GameState -> GameState)
+--
+(<@>) :: Reflex t => Behavior t (a -> b) -> Event t a -> Event t b
+(<@>) b = push $ \x -> do
+  f <- sample b
+  return . Just . f $ x
+
+infixl 4 <@
+
+-- | An version of '<@>' that does not use the value of the 'Event'.
+--
+-- Alternatively, it is 'tag' in operator form.
+--
+-- This is useful when we want to combine the values of several 
+-- 'Behavior's at particular points in time using an 'Applicative' 
+-- style syntax.
+--
+-- If we have:
+--
+-- > g  :: a -> b -> d
+-- > b1 :: Behavior t a
+-- > b2 :: Behavior t b
+-- > e  :: Event t c
+--
+-- where 'e' is firing at the points in time of interest.
+-- 
+-- Then we can use '<@':
+--
+-- > g <$> b1 <*> b2 <@  e :: Event t d
+--
+-- to combine the values of 'b1' and 'b2' at each of those points of time, 
+-- with the function 'g' being used to combine the values.
+--
+-- This is the same as '<@>' except that the 'Event' is being used only
+-- to act as a trigger. 
+(<@) :: (Reflex t) => Behavior t b -> Event t a -> Event t b
+(<@) = tag
 
 -- | A 'Monad' that supports adjustment over time.  After an action has been
 -- run, if the given events fire, it will adjust itself so that its net effect


spl:
Just saw this now and thought some feedback might be warranted, rather than let it languish without a response.

* The `Reflex t` constraint is not necessary in the `AttachApply` class declaration.
* Rather than have two functions, `attachApply` and `(<@>)`, that do the same thing, perhaps you should only have one. I guess that would be `(<@>)` based on your description.
* The name `AttachApply` doesn't seem to fit all instances, namely the `Identity` and `Maybe` instances. But I don't have a better suggestion.

It's an interesting concept. I'm not yet sure how useful it would be. Can you give more concrete examples of how you use it?

dalaing:
Thanks for having a look.

I'll definitely get rid of the `Reflex t` constraint on the class.  I'm not sure what I was thinking there.

I've taken the whole idea from `reactive-banana`, where I found it very useful.  In that setting they have `apply`, `<@>` and `<@`.  I'd be fine with not having `attachApply` in there and just having `<@>`.

I considered `Attachable` as a name, but it doesn't sit well with `Accumulator` (vs `Accumulatable`), so I figured `AttachApply` was a good place to start - I don't have any strong feelings about the name.

Back in `reactive-banana` land (and in `reflex` projects where I've added this change in via a Util module), it usually gets used with Applicative syntax to deal with transient combinations of several `Behavior`s and a sampling event.

So you have things like
```haskell
fWith3Args <$> bIn1 <*> bIn2 <@> eIn3
```
or
```haskell
fWith2Args <$> bIn1 <*> bIn2 <@ eJustATrigger
```

It's possibly to use `attachWith` for this (see [here](https://github.com/dalaing/reflex-sdl2/blob/master/examples/clicky/Main.hs#L143) for an example before and after the change), but it gets messy as the argument count increases.

It's also possible to define a data type for the set of behaviors that you want to combine, do something like
```haskell
MyDataType <$> bIn1 <*> bIn2
```
and go from there, but that's also a bit heavy weight when you just want to combine a handful of behaviors in one spot, when a particular event is firing.

At least when I'm coding (and having been exposed to these operators before), both of those approaches feel like they get in the way of me trying to tread lightly while decomposing my FRP code as much as I can.

The difference feels like going from using `>>=` and tuples to using `<*>`, which I probably shouldn't have been surprised by when I finally realised why these operators were in `reactive-banana`.

With all of that said, I'm happy to just keep them in a `Util` module somewhere if no-one else wants them :)

spl:
I don't have an opinion on whether it should go into `reflex`, but I appreciate the explanation. Now that I'm aware of it, I'll see if I can use it to improve my own code.

lynnard:
I think this can be quite useful - especially I think the applicative style looks clearer than having a `attachWith ...`.

I'd be happy to see this in; but I agree the `AttachApply` name feels a bit clunky.

src/Reflex/Class.hs
```diff
@@ -791,6 +795,64 @@ zipListWithEvent f l e = do
...
+-- > b1 :: Behavior t a
+-- > b2 :: Behavior t b
+-- > e  :: Event t c
+-- and an 'AttachApply' instance for `Behavior`, then we get:
+-- > f <$> b1 <*> b2 <@> e :: Event t d
+-- > g <$> b1 <*> b2 <@  e :: Event t d
+--
+-- It is a handy way to combine several 'Behavior's or 'Dynamic's together so that you can sample them all at the times when an `Event` is firing.
+class Reflex t => AttachApply t f where
+  attachApply :: f (a -> b) -> Event t a -> Event t b
```
ryantrinkle: This looks good, but we might want something like `attachApplyWithMaybeM :: (a -> b -> PushM t (Maybe c)) -> f a -> Event t b -> Event t c` to be in the typeclass, since it allows a wider range of functions to be to implemented with good performance.  Then, we can just define `attachApply` as `attachApplyWithMaybeM (\f x -> return $ Just $ f x)`.

src/Reflex/Class.hs
```diff
@@ -791,6 +795,64 @@ zipListWithEvent f l e = do
...
+instance Reflex t => AttachApply t Maybe where
+  attachApply (Just f) e = fmap f e
+  attachApply Nothing  _ = never
+
+instance Reflex t => AttachApply t (Behavior t) where
+  attachApply = attachWith ($)
+
+instance Reflex t => AttachApply t (Dynamic t) where
+  -- This is an inlined version of attachPromptyDynWith.
+  -- It is inlined to avoid having an orphan instance.
```
ryantrinkle: That's a very good point, but I think we should just move that function here, rather than repeating its body.

Also, I think it's worth considering whether the default operation ought to be attachPromptlyDynWith or a non-prompt version.  In the vast majority of code, I find that the prompt version is not necessary, and it make code more prone to loops.

ryantrinkle:
I might tend towards calling it `Sampleable`, `Variable`, or something like that.  Also, I think having an operator and a non-operator is fine - it's not too unusual in my experience.

@spl I'm not sure I understand the point about the Reflex superclass: since `Event t` is meaningless anyway without the Reflex constraint, it may end up being more convenient to have the superclass since users will sometimes be able to drop the Reflex constraint.  Is there something I'm overlooking?

dalaing:
I thought about something like `Sampleable`, but then would have wanted `sample` as the function name.  I'm OK with dropping the non-operator to have a nice class name without having to come up with a good function name to go with it :)

src/Reflex/Class.hs
dalaing: That's a good point - I'll look into the non-prompt version when I clean this up (hopefully tomorrow).

spl:
@ryantrinkle:

> I'm not sure I understand the point about the `Reflex` superclass: since `Event t` is meaningless anyway without the Reflex constraint, it may end up being more convenient to have the superclass since users will sometimes be able to drop the Reflex constraint. Is there something I'm overlooking?

The `Reflex t` constraint is needed on the instances anyway, so it doesn't simplify those. If you define a function using a combination of `attachApply` and another function with `Reflex t` (or some descendent thereof), then, yes, you may need to add a `Reflex t` constraint to the new function. The class constraint is not strictly necessary, but it also doesn't hurt.

src/Reflex/Class.hs
dalaing: From what I can see, this would also make a stronger case for `Reflex t` as a superclass of `AttachApply`, since it's needed to work with `PushM` regardless of which instance we're working with.

dalaing:
The type class is now renamed to `Sampelable`, and it's implemented in terms of `attachApplyWithMaybeM`.  

That seems to boost the convenience of having `Reflex t` as a superclass, but I did most of the most recent change while I was pre-coffee yesterday so I could be wrong about that :)

src/Reflex/Class.hs
```diff
@@ -821,6 +825,63 @@ zipListWithEvent f l e = do
...
+-- Once you have the instance that you need, you should probably be using it via the '<@>' and '<@' operators.
+class Reflex t => Sampleable t f where
+  attachApplyWithMaybeM :: (a -> b -> PushM t (Maybe c)) -> f a -> Event t b -> Event t c
+
+instance Reflex t => Sampleable t Identity where
+  attachApplyWithMaybeM f (Identity a) = push $ \x -> f a x
+
+instance Reflex t => Sampleable t Maybe where
+  attachApplyWithMaybeM f (Just a) = push $ \x -> f a x
+  attachApplyWithMaybeM _ Nothing  = const never
```
ryantrinkle: This instance strikes me as slightly dubious.  It relies heavily on the FunctorMaybe property of Events, whereas it seems that Sampleable may be susceptible to generalization beyond Events.  For example, imagine there were a type of Event with a fixed period (e.g. occurring once per second); it would not be in FunctorMaybe, but it would still totally make sense as something that could make use of a Sampleable.

ryantrinkle:
This is a really interesting interface, and I like <@> usage.  One thing I'm wondering about is whether we can avoid tying it to Events entirely - what if its characteristic function were, say, `MonadSample m => f a -> m a`.  I think that might give us even more power.  What do you think?

src/Reflex/Class.hs
dalaing: I can't think of many cases where I'd use the `Sampleable` class with the `Identity` or `Maybe` instances.  I generally just try to create instances for common types where there's only really one valid instance so that people don't have to create orphans or wrap things in newtypes.

I'm happy with getting rid of this if you don't want it, and if it's likely to interfere with future changes then it should definitely go.

dalaing:
This has parallels in another of the design space (from the perspective of someone coming from `reactive-banana` to `reflex`).  

One possibility for dealing with the introduction of `Dynamic` - although not one I'd necessarily recommend - would be to create an `EventLike` class with instances for `Event` and `Dynamic` and a `BehaviorLike` class with instances for `Behavior` and `Dynamic`.

It seems like that having a characteristic function with the signature that you suggest would be similar to making a break like that in conjunction with the decomposition of the main API into `MonadSample` and friends.

I like that idea, but I'd need to think about it for a while to be sure there weren't surprises there.  

It would be nice if the function could be called `sample` and work with either `Behavior` or `Dynamic`.

dalaing:
Just a thought here - and it's a bit embarrassing that I didn't start the PR off like this - maybe `<@>` and `<@` should both be written in terms of `Behavior`?

Considering how easy it is to get from `Dynamic` to `Behavior`, it seems like the typeclass might be overkill.

dalaing:
I've neatened this up and rebased it onto the `jsaddle` branch.

ryantrinkle:
@dalaing Looks great; thanks!

PR #70, opened 2016-10-18
Add distributeListOverDynPure, export distributeMapOverDynPure

spl:
I've found `distributeListOverDynPure` to be a very useful function that is non-trivial to write. I think it would be a good addition to `distributeDMapOverDynPure` and `distributeMapOverDynPure`. The latter was not exported and documented strangely (possibly a copy-paste error), so I exported it and updated the documentation.

### DIFF
--- a/src/Reflex/Dynamic.hs
+++ b/src/Reflex/Dynamic.hs
@@ -39,6 +39,8 @@ module Reflex.Dynamic
        , traceDynWith
        , splitDynPure
        , distributeDMapOverDynPure
+       , distributeMapOverDynPure
+       , distributeListOverDynPure
        , Demux
        , demux
        , demuxed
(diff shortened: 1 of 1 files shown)


mightybyte:
👍 I always wondered why we only had the DMap version of this!

ryantrinkle:
Thanks!

lynnard:
What's the difference between `distributeListOverDyn` in `Reflex.Class` vs `distributeListOverDynPure` in `Reflex.Dynamic`? (from what I see, there's no difference)

The `Pure` suffix seems a bit awkward given the already long name. I'd suggest change the name of `distributeMapOverDynPure` to just `distributeMapOverDyn` and move it into `Reflex.Class`; then all `distribute` functions will be in one place.

ryantrinkle:
Good point!  In general, the functions with 'Pure' in their name are ones that originally had a monadic version, and I am doing a deprecation cycle to get rid of the monadic version - so the Pure version is simply the non-Pure version, minus an invocation of `return`.  However, in this case, I think there's no actual deprecation cycle needed.

Issue #72, opened 2016-10-25
Default instance for Dynamic


mightybyte:
It would be nice to have a Default instance for Dynamic without having to define an orphan.  It would be the simple passthrough instance:

```
instance Default a => Dynamic t a where
    def = constDyn def
```

Is this something you would be open to?  It's not a huge problem, but would be nice to have when calling functions like this one https://github.com/reflex-frp/reflex-dom-semui/blob/master/src/Reflex/Dom/SemanticUI/Button.hs#L172 where you typically will be passing several `constDyn def` parameters.

ryantrinkle:
Yep, sounds good to me!

ryantrinkle:
Closing this in favor of #73.

PR #73, opened 2016-10-25
Add Default instance for Dynamic

mightybyte:
Here's a pull request for issue #72.  I wasn't sure if anything was needed in Spider.Internal, but GHC is happy with this so it seems not.


ryantrinkle:
Thanks!  Shouldn't need anything in Spider.

PR #77, opened 2016-11-20
removed stray bang pattern

mckeankylej:
Not sure how this is building in reflex-platform but this commit allows reflex to build standalone with the nix command:
```sh
nix-build -E "with (import <nixpkgs> {}); pkgs.haskell.packages.ghcjs.callPackage ./default.nix {}"
```


ryantrinkle:
Thanks!

PR #78, opened 2016-11-20
Move travis to use nix.

mckeankylej:
Travis now builds. This pr removes support for ghc 7.8.4 but adds support for ghcjs! The travis file is also significantly easier to read.


ryantrinkle:
Thanks!

PR #83, opened 2016-12-07
Add EventWriter

mightybyte:
I haven't had the chance to test this code yet (I have only proved it correct with GHC :P).  But I wanted to get it up to start discussion.

### DIFF
--- a/src/Reflex/EventWriter.hs
+++ b/src/Reflex/EventWriter.hs
@@ -42,7 +42,7 @@ import qualified Data.Some as Some
 import Data.Tuple
 import Data.Unique.Tag
 
--- | 'MonadEventWriter' efficiently collects 'Event' values using 'tellEvent'
+-- | 'EventWriter' efficiently collects 'Event' values using 'tellEvent'
 -- and combines them monoidally to provide an 'Event' result.
 class (Monad m, Monoid w) => EventWriter t w m | m -> t w where
   tellEvent :: Event t w -> m ()
--- a/src/Reflex/EventWriter.hs
+++ b/src/Reflex/EventWriter.hs
@@ -18,6 +18,8 @@ module Reflex.EventWriter
   ( EventWriterT (..)
   , runEventWriterT
   , EventWriter(..)
+  , runWithReplaceEventWriterTWith
+  , sequenceDMapWithAdjustEventWriterTWith
   ) where
 
 import Reflex.Class
@@ -33,10 +35,12 @@ import Control.Monad.Ref
 import Control.Monad.State.Strict
 import Data.Dependent.Map (DMap, DSum (..))
 import qualified Data.Dependent.Map as DMap
+import Data.Foldable
 import Data.Functor.Misc
 import qualified Data.List.NonEmpty as NonEmpty
 import Data.Maybe
 import Data.Semigroup
+import Data.Sequence
 import Data.Some (Some)
 import qualified Data.Some as Some
 import Data.Tuple
@@ -48,20 +52,17 @@ class (Monad m, Monoid w) => EventWriter t w m | m -> t w where
   tellEvent :: Event t w -> m ()
 
 -- | A basic implementation of 'EventWriter'.
-newtype EventWriterT t w m a = EventWriterT { unEventWriterT :: StateT [Event t w] m a }
+newtype EventWriterT t w m a = EventWriterT { unEventWriterT :: StateT (Seq (Event t w)) m a }
   deriving (Functor, Applicative, Monad, MonadFix, MonadIO, MonadException)
 
 -- | Run a 'EventWriterT' action.
-runEventWriterT
-    :: (Reflex t, Monad m, Semigroup w)
-    => EventWriterT t w m a
-    -> m (a, Event t w)
+runEventWriterT :: (Reflex t, Monad m, Semigroup w) => EventWriterT t w m a -> m (a, Event t w)
 runEventWriterT (EventWriterT a) = do
   (result, requests) <- runStateT a mempty
-  return (result, mconcat $ reverse requests)
+  return (result, mconcat $ toList requests)
 
 instance (Reflex t, Monad m, Monoid w) => EventWriter t w (EventWriterT t w m) where
-  tellEvent w = EventWriterT $ modify (w:)
+  tellEvent w = EventWriterT $ modify (|> w)
 
 instance MonadTrans (EventWriterT t w) where
   lift = EventWriterT . lift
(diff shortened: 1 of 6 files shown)


src/Reflex/EventWriter.hs
```diff
@@ -0,0 +1,172 @@
...
+import Data.Maybe
+import Data.Semigroup
+import Data.Some (Some)
+import qualified Data.Some as Some
+import Data.Tuple
+import Data.Unique.Tag
+
+-- | 'MonadEventWriter' efficiently collects 'Event' values using 'tellEvent'
+-- and combines them monoidally to provide an 'Event' result.
+class (Monad m, Monoid w) => EventWriter t w m | m -> t w where
```
spl: The class is named `EventWriter` but the comment refers to it as `MonadEventWriter`.

src/Reflex/EventWriter.hs
```diff
@@ -0,0 +1,172 @@
...
+  deriving (Functor, Applicative, Monad, MonadFix, MonadIO, MonadException)
+
+-- | Run a 'EventWriterT' action.
+runEventWriterT
+    :: (Reflex t, Monad m, Semigroup w)
+    => EventWriterT t w m a
+    -> m (a, Event t w)
+runEventWriterT (EventWriterT a) = do
+  (result, requests) <- runStateT a mempty
+  return (result, mconcat $ reverse requests)
```
spl: Is the `reverse` important? If so, perhaps something other than a list should be used. Alternatives include a `Seq` or just one `Event  t w` in the `StateT` with `tellEvent` using `mappend` in the appropriate order.

spl:
Not bad. The idea, as I understand it, is to reduce the work of tracking a number of `Event`s in code only to `leftmost` them all in the end.

cgibbard:
The main hesitation with something like this (which extends to
DynamicWriter as well) is that while there are cases where it seems
appropriate, it is somewhat antithetical to the primary goal of FRP which
is to make it easy (or at least easier) to determine what can affect any
given value.

There's a lot to be said for the ability to answer the question of "why is
this Event firing?" or generally "what affects the state of this thing?"
with "let's look at its definition". When that definition gets spread out
throughout your entire codebase, you have sort of incurred the cost that
people using global IORefs have: it could be affected from anywhere, and it
can become difficult to know which 'tell' is responsible.

Nevertheless, there are still cases where one way or the other, you have
some Event which will ultimately need to be triggered from almost
everywhere in your application, and it can get tedious explicitly passing
parts of it upward to be merged. A counterpoint to that is that while it
may indeed be annoying, it's an annoyance which spreads awareness of this
complex event to the entire development team, and helps ensure that the
parts of the code which really can't affect it won't claim to be able to.

Even with all the caveats, I think there will still be some cases where
it's nonetheless appropriate to use this.

So it's to some degree a question of how can we educate people to be really
quite certain that surgery of this sort is making their life better rather
than incurring risks of complication without sufficient benefit?

On Wed, Dec 7, 2016, 02:51 Sean Leather <notifications@github.com> wrote:

Not bad. The idea, as I understand it, is to reduce the work of tracking a
number of Events in code only to leftmost them all in the end.

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/reflex-frp/reflex/pull/83#issuecomment-265379752>, or mute
the thread

spl:
@cgibbard It sounds to me like you are describing the general situation (e.g. the pros and cons) of using state or things like `StateT`. In other words, your concern is not specifically about `EventWriter` as presented here but about the idea of carrying implicit values in code and the discoverability of the origins of those values. While that is certainly a valid concern, is it appropriate for this issue? Are you trying to say something like “`reflex` should not have this feature because it potentially promotes bad code”? (This is, of course, a reasonable argument.) Or are we going too far offtrack? (I'm not advocating one way or another. Just trying to focus the thread. 😉)

cgibbard:
Something like that. Reflex specifically exists in the first place as a principled attack on untangling the complexity of managing interconnected networks of things that change over time. Introducing facilities of this sort, which allow for not-explicitly-related bits of code to interact, in a sense partially inverts what Reflex is set up to do in the first place.

The concern is that if you put code in a library, people will find it and use it, often before really understanding what effect it might have in the large on the maintainability of their program. These writers-of-FRP-values need perhaps more experience to derive benefit from than most people have at the moment, and while there almost certainly are legitimate use cases for them, they look far more innocuous than they really are (especially to a potential beginner who is comfortable with the usual industry practice of having everything affect everything else all the time ;) ).

The stated purpose of the PR is to provoke discussion, so I don't think it's too far off-track for us to discuss what the ecosystem-level impact of having things like this around might be, and how we're going to manage that.

spl:
@cgibbard Thanks for the clarification. 😄

src/Reflex/EventWriter.hs
mightybyte: Good catch, thanks!

paolino:
I'd like to give my two cents just to understand better and maybe just to open the discussion to what we want to have as a library. Reflex is a good Library for building Networks and it's done so that we have a deterministic effect  for events going through it. This is ruling out cases where a we are only interested in the communication part. There are cases we are not interested in synchronicity. it's what it's been said we need a kind of a wireless communication, where reflex is not supposed to be used for building network but the more about making possible  asynchronous communication. if this is what we want probably acting with eventswriter is using a synchronous tool for something not interested in the synchronicity.

src/Reflex/EventWriter.hs
treeowl: Reversing shouldn't be at all hard to work around. Ignoring the hopeless case of an infinite list,

```haskell
mconcat $ reverse requests
= foldr mappend mempty (reverse requests)  -- Monoid class law
= foldl mappend mempty (reverse requests) -- Associativity of mappend, identity property of mempty, and finiteness of the list
= foldr (flip mappend) mempty requests -- foldl/reverse relationship
```

So we can just use that instead, if reversing is important.

ryantrinkle: Except, for Events, mconcat is *far* more performant than a series of mappends.  Using Seq or something would make sense.

To answer @spl, yes, I think the reverse is important, so that the monoid is used in the same way that it is with WriterT.  I wouldn't want someone to end up surprised by things coming out in the reverse order!

treeowl: @ryantrinkle, that still doesn't mean you have to reverse the list!

```haskell
foldr (flip mappend) mempty requests
= getDual $ foldr mappend mempty (map Dual requests) -- Monoid instance for Dual
= getDual $ mconcat (map Dual requests) -- mconcat law
= getDual $ mconcat (coerce requests) -- map/coerce law
```

treeowl: Sorry--I missed something

```haskell
getDual $ mconcat (map Dual requests) -- mconcat law
= getDual $ mconcat (map (fmap Dual) requests) -- Monoid instance for Event
```

So you may not get `map/coerce` magic after all, but you don't need `reverse`.

treeowl: May not matter in the end.

PR #85, opened 2017-01-25
Reduce the number of cache events in basic Reflex code; expose cheap …

danharaj:
…primitives in Reflex.Class; add a test for a GC bug that was discovered in the process of writing this patch

### DIFF
--- a/src/Reflex/Class.hs
+++ b/src/Reflex/Class.hs
@@ -234,7 +234,7 @@ class ( MonadHold t (PushM t)
   -- 'Behavior's and hold 'Event's, and use the results to produce a occurring
   -- (Just) or non-occurring (Nothing) result
   push :: (a -> PushM t (Maybe b)) -> Event t a -> Event t b
-  -- | Like 'push' but intended for functions that the implementation can consider cheap to compute for performance considerations.
+  -- | Like 'push' but intended for functions that the implementation can consider cheap to compute for performance considerations. WARNING: The function passed to 'pushCheap' may be run multiple times without any caching.
   pushCheap :: (a -> PushM t (Maybe b)) -> Event t a -> Event t b
   -- | Create a 'Behavior' by reading from other 'Behavior's; the result will be
   -- recomputed whenever any of the read 'Behavior's changes
--- a/src/Reflex/Class.hs
+++ b/src/Reflex/Class.hs
@@ -97,6 +97,14 @@ module Reflex.Class
   , appendEvents
   , onceE
   , sequenceThese
+  , fmapMaybeCheap
+  , fmapCheap
+  , fforCheap
+  , fforMaybeCheap
+  , pushAlwaysCheap
+  , tagCheap
+  , mergeWithCheap
+  , mergeWithCheap'
 #ifdef SPECIALIZE_TO_SPIDERTIMELINE_GLOBAL
   , Spider
   , SpiderEnv
@@ -109,6 +117,7 @@ module Reflex.Class
   , never
   , constant
   , push
+  , pushCheap
   , pull
   , merge
   , fan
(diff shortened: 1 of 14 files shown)


src/Reflex/Class.hs
```diff
@@ -225,6 +234,8 @@ class ( MonadHold t (PushM t)
   -- 'Behavior's and hold 'Event's, and use the results to produce a occurring
   -- (Just) or non-occurring (Nothing) result
   push :: (a -> PushM t (Maybe b)) -> Event t a -> Event t b
+  -- | Like 'push' but intended for functions that the implementation can consider cheap to compute for performance considerations.
```
ryantrinkle: Let's put a bigger, scarier comment on this.  We need to make it extremely clear that the user's function may be run multiple times.

src/Reflex/Patch.hs
```diff
@@ -5,11 +5,14 @@
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE StandaloneDeriving #-}
 {-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
 module Reflex.Patch
   ( Patch (..)
   , PatchDMap (..)
   , ComposeMaybe (..)
   , PatchMap (..)
+  , Group (..), Additive
```
ryantrinkle: Should `Additive` export `(..)`?  Also, should probably be on the next line.

src/Reflex/Patch.hs
```diff
@@ -95,3 +86,18 @@ instance Ord k => Semigroup (PatchMap k v) where
 instance Ord k => Monoid (PatchMap k v) where
   mempty = PatchMap mempty
   mappend = (<>)
+
+-- Patches based on commutative groups
```
ryantrinkle: Haddock syntax?

src/Reflex/Pure.hs
```diff
@@ -66,6 +66,8 @@ instance (Enum t, HasTrie t, Ord t) => Reflex (Pure t) where
   push :: (a -> PushM (Pure t) (Maybe b)) -> Event (Pure t) a -> Event (Pure t) b
   push f e = Event $ memo $ \t -> unEvent e t >>= \o -> f o t
 
+  pushCheap = push
```
ryantrinkle: Type signature.

src/Reflex/Spider/Internal.hs
```diff
@@ -1641,9 +1666,12 @@ mergeCheap d = Event $ \sub -> do
...
           , subscriberRecalculateHeight = \_ -> return ()
           }
-    (EventSubscription _ changeSubd, change) <- subscribeAndRead (dynamicUpdated d) s
+    (EventSubscription sln changeSubscription, change) <- subscribeAndRead (dynamicUpdated d) s
     forM_ change $ \c -> defer $ SomeMergeUpdate (updateMe c) invalidateMyHeight revalidateMyHeight
-    liftIO $ writeIORef changeSubdRef (s, changeSubd)
+    -- We explicitly hold on to the unsubscribe function from subscribing to the update event.
+    -- If we don't do this, there are certain cases where mergeCheap will fail to properly retain
+    -- its subscription.
+    liftIO $ writeIORef changeSubdRef (s, sln, changeSubscription)
```
ryantrinkle: Let's not unpack and re-pack the EventSubscription, in line with the new comment above.

src/Reflex/Spider/Internal.hs
```diff
@@ -1935,7 +1965,7 @@ instance HasSpiderTimeline x => R.Reflex (SpiderTimeline x) where
   {-# INLINABLE current #-}
   current = SpiderBehavior . dynamicCurrent . unSpiderDynamic
   {-# INLINABLE updated #-}
-  updated = SpiderEvent . fmap runIdentity . dynamicUpdated . unSpiderDynamic
+  updated = coerceWith Coercion $ SpiderEvent . dynamicUpdated . unSpiderDynamic
```
ryantrinkle: How is `coerceWith Coercion` different from `coerce`?

src/Reflex/Patch.hs
danharaj: If you export with (..) for a class that doesn't have any members you get a warning. I can put it on another line tho.

ryantrinkle: Ah, ok; cool!

PR #88, opened 2017-02-01
Clarify difference between Incremental and Dynamic

eskimor:
Btw. do you have any code at hand making use of Incrementals, I get a feeling it could help simplify some code of mine, but a usage example would really help.

### DIFF
--- a/src/Reflex/Class.hs
+++ b/src/Reflex/Class.hs
@@ -211,13 +211,15 @@ class ( MonadHold t (PushM t)
   -- occurring or not occurring; if it is occurring, it will contain a value of
   -- the given type (its "occurrence type")
   data Event t :: * -> *
-  -- | A 'Dynamic' is just an 'Incremental' whose patch type always completely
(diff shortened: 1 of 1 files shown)


ryantrinkle:
You can see a use of an incremental [here](https://github.com/reflex-frp/reflex/blob/develop/src/Reflex/EventWriter.hs#L133), but I'm afraid it's not likely to simplify anything until we've got a much nicer set of combinators for Incrementals; right now, Incrementals are pretty much only used for performance-critical code, via mergeIncremental.  The primary use case is to make small changes to a large merge.

ryantrinkle:
I think we should probably do this documentation change the other way around - most users shouldn't need to understand what an Incremental is at all, so I don't want to send users trying to understand Dynamic over to Incremental first.  Do you think you could reword this, perhaps to explain in Incremental's docs that it's a generalization of Dynamic?

eskimor:
Thank you!
Better now?

ryantrinkle:
Looks great, thanks!

PR #101, opened 2017-04-04
haskell-src-exts 1.18 support

hamishmack:
Also fixes for hlint and tell nix to ignore `dist` directory


ryantrinkle:
Awesome; thanks!

PR #105, opened 2017-04-25
TriggerEventT improvements

ElvishJerricco:
- Made `EventTriggerRef` concrete in `IORef`
- Added a `PrimMonad` instance
- Added a `MonadAdjust` instance


src/Reflex/TriggerEvent/Base.hs
```diff
@@ -5,6 +5,7 @@
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE TypeFamilies #-}
+
```
ryantrinkle: Reasonable, but should be done globally, not just in this one source file.  Also, I think a module comment would serve better.

src/Reflex/Class.hs
```diff
@@ -999,7 +999,7 @@ class (Reflex t, Monad m) => MonadAdjust t m | m -> t where
   traverseDMapWithKeyWithAdjust :: GCompare k => (forall a. k a -> v a -> m (v' a)) -> DMap k v -> Event t (PatchDMap k v) -> m (DMap k v', Event t (PatchDMap k v'))
   traverseDMapWithKeyWithAdjustWithMove :: GCompare k => (forall a. k a -> v a -> m (v' a)) -> DMap k v -> Event t (PatchDMapWithMove k v) -> m (DMap k v', Event t (PatchDMapWithMove k v'))
 
-instance (Reflex t, MonadAdjust t m) => MonadAdjust t (ReaderT r m) where
+instance MonadAdjust t m => MonadAdjust t (ReaderT r m) where
```
ryantrinkle: Looks good; i wonder why this wasn't a warning; GHC bug?

src/Reflex/TriggerEvent/Base.hs
```diff
@@ -29,21 +33,28 @@ import Reflex.TriggerEvent.Class
 data TriggerInvocation a = TriggerInvocation a (IO ())
 
 -- | A reference to an 'EventTrigger' suitable for firing with 'TriggerEventT'.
-newtype EventTriggerRef t m a = EventTriggerRef { unEventTriggerRef :: Ref m (Maybe (EventTrigger t a)) }
+newtype EventTriggerRef t a = EventTriggerRef { unEventTriggerRef :: IORef (Maybe (EventTrigger t a)) }
```
ryantrinkle: What's the benefit of monomorphizing this?

ElvishJerricco: `m` means a lot of different things in a lot of different contexts. In `ImmediateDomBuilderT`, `m` would mean the base monad. But elsewhere, you want to be able to use `DomBuilder t m => ... -> EventTriggerRef t m -> ...` or something. If we want it to be polymorphic, it should just be polymorphic over the `Ref` type, not `m` itself.

src/Reflex/TriggerEvent/Base.hs
ElvishJerricco: Oh right. Force of habit. I'll add a commit removing this.

src/Reflex/TriggerEvent/Base.hs
ryantrinkle: That's a good point.  I'm mostly thinking that we might want to support STRef, TVar, or something else like that at some point.

src/Reflex/TriggerEvent/Base.hs
ryantrinkle: NP

ElvishJerricco:
Are we expecting the travis to fail? The error is pretty strange.

ryantrinkle:
Travis seems pretty unreliable here.  I don't know what's up, and haven't had a chance to debug it.

src/Reflex/TriggerEvent/Base.hs
ElvishJerricco: I don't think it'll be hard to change when that time comes. `IORef` is already hard coded in a number of places.

PR #106, opened 2017-04-26
Replace and deprecate HList, FHList, distributeFHListOverDynPure and collectDynPure.

adamConnerSax:
This replaces/deprecates HList and FHList with the NP type from generics-sop.  It has equivalent functionality and much greater support from the generics-sop library.  It replaces/deprecates dsitributeFHListOverDynPure and collectDynPure with equivalent functionality built from NP and replaces the use of distributeFHListOverDynPure in qDynPure (in Reflex.Dynamic.TH) with the NP version.
This version of collectDynPure is more powerful, or at least easier to deploy, than the previous one since it requires only that the types be instances of generic-sop's Generic type (easily derived from the GHC Generic type), rather than needing to write HList and FHList instances. 
It adds tests of this functionality: a standalone Pure-only test comparing mkDynPure quasiquoting with zipDynWith and a pure vs spider comparison of mkDynPure in the semantics test suite. More testing might be useful. 
It adds some supporting functions for converting generics-sop NP to dependent maps and back.  This should properly be in a separate library but until that is done, it is placed here, in Generics.SOP.DMapUtilties.
This adds one library dependency, namely on generics-sop (>=0.2.4.0).
Older types and functions have been retained but deprecated.


src/Generics/SOP/DMapUtilities.hs
```diff
@@ -0,0 +1,174 @@
...
+import Data.GADT.Compare ((:~:) (..), GCompare (..), GEq (..), GOrdering (..))
+
+
+import Data.Functor.Identity (Identity (runIdentity))
+
+
+-- |A Tag type for making a 'DM.DMap' keyed by a type-level list
+data TypeListTag (xs :: [k]) (x :: k) where -- x is in xs
+  Here  :: TypeListTag (x ': xs) x          -- x begins xs
+  There :: TypeListTag xs x -> TypeListTag (y ': xs) x -- given that x is in xs, x is also in (y ': xs)
```
ryantrinkle: `Here` and `There` are very nearly a naming conflict with `here` and `there` in `Data.These`.  Are these names chosen to be consistent with something else, or would it be possible to change them to avoid this near-conflict?

src/Reflex/Dynamic/FactorDynGeneric.hs
```diff
@@ -0,0 +1,90 @@
...
+      nsnpDyn' = hcmap sListIC (hmap (Comp. I) . distributeI_NP . getCompose) <$> nsnpDyn -- Dynamic t (NS (NP (I :.: Dynamic t)) (Code a))
+      nsnpDyn'' = nsOfnpUnCompose <$> nsnpDyn' -- Dynamic t (NS (NP I) (Code b))
+      result = to . SOP <$> nsnpDyn''
+  return result
+
+-- for example:
+maybeDyn :: (Reflex t, MonadHold t m, MonadFix m) => Dynamic t (Maybe a) -> m (Dynamic t (Maybe (Dynamic t a)))
+maybeDyn = factorDynGeneric
+
+eitherDyn :: (Reflex t, MonadHold t m, MonadFix m) => Dynamic t (Either a b) -> m (Dynamic t (Either (Dynamic t a) (Dynamic t b)))
```
ryantrinkle: This is really cool!

ryantrinkle:
This looks really awesome!  I've updated it on the [generics-sop branch](https://github.com/reflex-frp/reflex/tree/generics-sop) to the latest reflex develop.  However, it looks like generics-sop uses a lot of template-haskell, which means we can't support this on mobile :(  Currently, everything in reflex supports mobile fully, so we'll need to think about how to approach this.  I'd really like to get this in!

src/Reflex/Dynamic/FactorDynGeneric.hs
adamConnerSax: Thanks!  I still think there may be a bug but though I first found it a while ago, I haven't had a chance to go back and update the reflex version (maybe I'll try that now) or make a simpler case (I found it amidst all the profunctor stuff).  But we should definitely make sure there are a few tests.

adamConnerSax:
I'd be happy to help get it in.  I've got no idea how to work around the TH issue for mobile.  It might be the case that it only interferes with factorDyn and not all the HList replacement stuff since it only comes into generics-sop for deriving generic instances.  So it doesn't get used for the NP and NS types and so the HList replacement stuff might all be okay.  But I have to look again.

ryantrinkle:
@adamConnerSax I just pushed a merged version of this here: https://github.com/reflex-frp/reflex/tree/factorDynGeneric-update

However, until https://github.com/well-typed/generics-sop/issues/61 is fixed, it sounds like we probably can't use generics-sop.

ali-abrar:
It looks like generics-sop was split up so that there's a no-TH sop-core package. That package doesn't appear to have everything we'd need to get this branch working without TH, though.

ali-abrar:
But, I've been reminded that we don't need to avoid TH anymore anyhow.

ali-abrar:
I'm working on updating this so that we can merge it.

Ericson2314:
We don't need non-TH anymore I don't think

PR #111, opened 2017-05-30
Add AppendMap.

luigy:


reflex.cabal
```diff
@@ -69,6 +69,7 @@ library
     transformers-compat >= 0.3
 
   exposed-modules:
+    Data.AppendMap
```
ryantrinkle: Does this need a comma?

src/Data/AppendMap.hs
```diff
@@ -0,0 +1,417 @@
...
+  def = empty
+
+instance AppendMap k' v' ~ t => Rewrapped (AppendMap k a) t
+
+instance Wrapped (AppendMap k a) where
+  type Unwrapped (AppendMap k a) = Map k a
+  _Wrapped' = iso (\(AppendMap m) -> m) AppendMap
+
+-- | Operator for creating a singleton 'Map'
+(=:) :: k -> a -> AppendMap k a
```
ryantrinkle: Do you think this will conflict with the other (=:) that we have?

reflex.cabal
luigy: Good catch! I'm surprised cabal didn't complain about this

src/Data/AppendMap.hs
luigy: Yes, but currently this is not exported by `Reflex.Dom`. Either way probably best to drop this to avoid any clashes in the future.

ryantrinkle:
Thanks!

k-bx:
There's http://hackage.haskell.org/package/monoidal-containers available, which looks similar. Would using that one be a good fit maybe?

ElvishJerricco:
Hm. Close, but I think that package would need some extra work. It doesn't support nearly as many of the Map functions, and the constraint on the values doesn't need to be `Monoid`, just `Semigroup`.

PR #114, opened 2017-06-11
Add a widget module; add fast sortable list widget

### LINKED ISSUE #113: Upstream `simpleSortableList`
Upstream `simpleSortableList` from [here](https://github.com/reflex-frp/reflex-dom/blob/develop/reflex-dom/examples/sortableList.hs#L70) and add documentation.

```haskell
-- | Provides a framework for building sortable content in such a way that resorting it can
-- cause minimal disruption. For example, naively resorting a list of images would destroy
-- every image and add them back in the new order. This framework is able to avoid that by
-- preserving the identity of each image and simply moving it to the new location.
simpleSortableList :: forall t m k v. (MonadHold t m, MonadFix m, MonadAdjust t m, Ord k)
                   => (k -> v -> m ()) -- ^ A function to render the content for a key/value in a @Map@
                   -> Map k v -- ^ A @Map@ of values identified by keys
                   -> Event t (v -> v -> Ordering) -- ^ An event carrying a sort function that can resort without destroying/rebuilding each element
                   -> Event t (v -> v -> Ordering) -- ^ An event carrying a sort function that requires a resorting to rebuild each element
                   -> m ()
simpleSortableList f m0 resortFunc resortSlowFunc = do
  rec let resortPatchFast = attachWith (flip patchThatSortsMapWith) (currentIncremental m) resortFunc
          redrawPatch :: Map k v -> (v -> v -> Ordering) -> PatchMapWithMove k v
          redrawPatch d cmp = unsafePatchMapWithMove $ fmap (MapEdit_Insert False) $ Map.fromList $ zip (Map.keys d) (sortBy cmp $ Map.elems d)
          resortPatchSlow = attachWith redrawPatch (currentIncremental m) resortSlowFunc
          resortPatch = leftmost
            [ resortPatchFast
            , resortPatchSlow
            ]
      m <- holdIncremental m0 resortPatch
  _ <- mapMapWithAdjustWithMove f m0 resortPatch
  return ()
```

3noch:
Fixes #113

### TODO:

  - [x] Do we need the event for slow re-sorting? I see why it was used in the original example, but I don't see it's usefulness in any other case.
  - [x] Test
  - [x] Add tests to a test suite? (I'm not sure what you do normally)

Also I'm curious if it'd be worth providing a version of this that accepts `Dynamic t (Map k v)` and handles inserts and deletes as well.

### DIFF
--- a/reflex.cabal
+++ b/reflex.cabal
@@ -97,7 +97,8 @@ library
     Reflex.Requester.Base,
     Reflex.Requester.Class,
     Reflex.Spider,
-    Reflex.Spider.Internal
+    Reflex.Spider.Internal,
+    Reflex.Widget.Basic
 
   ghc-options: -Wall -fwarn-tabs -funbox-strict-fields -O2
   ghc-prof-options: -fprof-auto
--- a/src/Reflex/Widget/Basic.hs
+++ b/src/Reflex/Widget/Basic.hs
@@ -1,0 +1,42 @@
+{-# LANGUAGE CPP #-}
+{-# LANGUAGE ExplicitForAll #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE RecursiveDo #-}
+#ifdef USE_REFLEX_OPTIMIZER
+{-# OPTIONS_GHC -fplugin=Reflex.Optimizer #-}
+#endif
+
+module Reflex.Widget.Basic where
+
+import Control.Monad.Fix (MonadFix)
+import Data.Map (Map)
+import Reflex.Class
+import Reflex.Patch.MapWithMove
+
+
+-- | Build sortable content in such a way that re-sorting it can cause minimal
+-- disruption to an existing context.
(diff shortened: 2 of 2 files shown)


3noch:
@ryantrinkle Build is still failing but not because of this code, AFAIC.

ryantrinkle:
@3noch Definitely don't need slow version; that was just so I could demonstrate the difference in performance.

Don't worry about a test for now.  We do need to get more unit testing going - and headless chrome and StaticDomBuilderT should make this much easier - but we don't really have a framework in place currently.

3noch:
@ryantrinkle Ok took that out. That certainly simplified things!

3noch:
@ryantrinkle I just used this in a project and it works!

3noch:
@ryantrinkle I updated this to keep the results of the dom builders for each list item.

ryantrinkle:
Awesome, thanks!

Issue #99, opened 2017-03-27
Installing reflex-0.5.0 fails with "Kind incompatibility when matching types" in Patch.hs


hansroland:
GHCJS Version: 0.2.0 (GHC 7.10.3).
GHCJS Download date: 2017-03-27.
I deleted directory /home/roland/.ghcjs and re-installed the GHCJS compiler.
Clone date of https://github.com/reflex-frp/reflex: 2017-03-27.

File  [reflex-ghcjs.txt](https://github.com/reflex-frp/reflex/files/872397/reflex-ghcjs.txt) contains the full installation log.

**Error:**

src/Reflex/Patch.hs:106:45:
    Kind incompatibility when matching types:
      g0 :: k -> *
      Constant () :: * -> *
    Expected type: k1 v1 -> ComposeMaybe v v1 -> Maybe (g0 v1)
      Actual type: k1 v1 -> ComposeMaybe v v1 -> Maybe (Constant () b0)
    Relevant bindings include
      deletions :: DMap k1 g0 (bound at src/Reflex/Patch.hs:106:11)
      old :: PatchTarget (PatchDMap k1 v)
        (bound at src/Reflex/Patch.hs:104:26)
      diff :: DMap k1 (ComposeMaybe v)
        (bound at src/Reflex/Patch.hs:104:20)
      apply :: PatchDMap k1 v
               -> PatchTarget (PatchDMap k1 v)
               -> Maybe (PatchTarget (PatchDMap k1 v))
        (bound at src/Reflex/Patch.hs:104:3)
    In the first argument of ‘DMap.mapMaybeWithKey’, namely
      ‘(const $ nothingToJust . getComposeMaybe)’
    In the expression:
      DMap.mapMaybeWithKey (const $ nothingToJust . getComposeMaybe) diff
    In an equation for ‘deletions’:
        deletions
          = DMap.mapMaybeWithKey
              (const $ nothingToJust . getComposeMaybe) diff

src/Reflex/Patch.hs:384:79:
    The second argument of ‘Constant’ should have kind ‘*’,
      but ‘a’ has kind ‘k’
    In the type signature for ‘deleteFunc’:
      deleteFunc :: forall a.
                    k a -> DMapEdit k v a -> Maybe (Constant () a)
    In an equation for ‘apply’:
        apply (PatchDMapWithMove p) old
          = Just $! insertions `DMap.union` (old `DMap.difference` deletions)
          where
[...]

PR #115, opened 2017-06-12
Specify 8.0.1 as minimal GHC release

hansroland:
It's no longer possible to compile reflex with GHC-7.0.3.  The current error is:

```
src/Data/AppendMap.hs:10:14:
    Unsupported extension: TypeApplications
```
This PR sets the necessary version of base to 4.9 so the minimal GHC version is now 8.0.1. 

After merge of this PR, issue #99 can be closed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,6 @@ cache:
     - /nix
 matrix:
   include:
-  - env: COMPILER=ghc7103
   - env: COMPILER=ghc801
   - env: COMPILER=ghcjs
 script:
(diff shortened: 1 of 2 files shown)


ElvishJerricco:
I think it'd probably be bettter to remove the dependency on GHC 8. TypeApplications should be easy to avoid.

hansroland:
@ElvishJerricco: As you can see in issue 99 there are more things to fix than TypeApplications. I used to have a [stack build version based on lts-6.25](https://github.com/hansroland/reflex-dom-inbits/blob/bb6637d4dffba2fea5388a6d05322d6afe03c5fb/stack.yaml) which no longer works. I don't think it makes sense to invest time into 7.10.3

hansroland:
Of course, I prefer fixes #116 and #117.

PR #116, opened 2017-06-12
Eliminate the only use of TypeApplications

cgibbard:
We should be able to retain support for older GHCs a bit longer this way. (But see issue 99 for another problem facing 7.10.x.)


lynnard:
Do you know a way to address #99? I think that depends on ghc 8 as well.

luigy:
@lynnard I opened https://github.com/reflex-frp/reflex/pull/117 which should fix 7.10 build. Lmk if that works for you

PR #121, opened 2017-07-08
Renames MonadAdjust and moves it to its own module.

dalaing:
This also brings the collection management functions across from `reflex-dom` - there is a companion PR in `reflex-dom` dealing with this [here](https://github.com/reflex-frp/reflex-dom/pull/163).

### DIFF
--- a/reflex.cabal
+++ b/reflex.cabal
@@ -79,12 +79,14 @@ library
 
   exposed-modules:
     Data.AppendMap,
+    Data.Map.Misc,
     Data.Functor.Misc,
     Data.WeakBag,
     Data.FastWeakBag,
     Data.FastMutableIntMap,
     Reflex,
     Reflex.Class,
+    Reflex.Collection,
     Reflex.EventWriter,
     Reflex.Dynamic,
     Reflex.Dynamic.Uniq,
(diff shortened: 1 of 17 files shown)


ryantrinkle:
Let's do something like this: https://gist.github.com/ryantrinkle/82ba8e95cc2ddd0bdf2c54c7317c645c

dalaing:
I'm trying to work out how to do that without a cyclic import :/

dalaing:
I might move Adjustable back into Reflex.Class, but have the list function in Reflex.Adjustable.Base? or something like that?

dalaing:
The current plan is to move `Adjustable` back into `Reflex.Class` and to move the collection management functions into `Reflex.Collection` - if I'm doing that, should I move `sortableList` from `Reflex.Widget.Basic` into `Reflex.Collection` as well?

ryantrinkle:
Hmm, maybe we should bite the bullet and break which module it's in.  As long as we make sure to re-export from Reflex, most people probably wan't hit it (I think importing Reflex.Class itself directly is probably not super-common).  I don't know how to do a smooth transition for the actual move.

PR #127, opened 2017-08-15
Brings in the Time module from `reflex-dom`

dalaing:
I haven't tested this yet, but I'm pushing what I have so that I don't forget that I have it a branch somewhere. There is a companion PR for `reflex-dom` [here](https://github.com/reflex-frp/reflex-dom/pull/168).


ryantrinkle:
Thanks!

PR #128, opened 2017-08-31
Add convenience function fmapEither

samtay:
I'm seeing variants of this all over the place. What do you think about adding it as a convenience? Do you think instead we should have corresponding `fmapLeft` and `fmapRight`, as I've seen both in a number of places, but `Right` is more common.

### DIFF
--- a/src/Reflex/Class.hs
+++ b/src/Reflex/Class.hs
@@ -100,6 +100,8 @@ module Reflex.Class
   , FunctorMaybe (..)
   , fforMaybe
   , ffilter
+  , filterLeft
+  , filterRight
     -- * Miscellaneous convenience functions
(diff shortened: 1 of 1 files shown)


ryantrinkle:
Great idea!  Let's go with `fmapLeft` and `fmapRight`, but other than that, this looks good to me.

samtay:
Hmm. I went with `filter*` instead of `fmap*` prefixes before seeing your comment. I think this makes more sense, because with `fmapMaybe` you are still lifting a supplied function from the caller, whereas these functions don't lift any user supplied functions and are just filtering values already in a `FunctorMaybe` context.

ryantrinkle:
Thanks!

PR #134, opened 2017-10-03
Adds withEventWriterT.

dalaing:


dalaing:
I'm wondering if it would be nice to have something in the `EventWriter` class to support the use of this further down a transformer stack (although I'm not sure off the top of my head if that is doable or not).

ryantrinkle:
Thanks!

PR #138, opened 2017-11-02
Moves Workflow across from reflex-dom

dalaing:
This also included moving generalizations of `dyn` and `widgetHold` across from `reflex-dom` and breaking out a `NotReady` class from the `DomBuilder` class.

There is a companion PR in the `reflex-dom` repository [here](https://github.com/reflex-frp/reflex-dom/pull/178).

### DIFF
--- a/src/Reflex/Workflow.hs
+++ b/src/Reflex/Workflow.hs
@@ -1,0 +1,37 @@
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE RecursiveDo #-}
+module Reflex.Workflow (
+  -- * Workflows
+    Workflow (..)
+  , workflow
+  , workflowView
+  , mapWorkflow
+  , mapWorkflowCheap
+  ) where
+
+import Control.Arrow ((***))
+import Control.Monad.Fix (MonadFix)
+
+import Reflex.Class
+import Reflex.Network
+import Reflex.NotReady.Class
+import Reflex.PostBuild.Class
+
+newtype Workflow t m a = Workflow { unWorkflow :: m (a, Event t (Workflow t m a)) }
+
+workflow :: forall t m a. (Reflex t, Adjustable t m, MonadFix m, MonadHold t m) => Workflow t m a -> m (Dynamic t a)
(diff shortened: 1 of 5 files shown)


dalaing:
I'm now wonder if NotReady might be doable as a few extra methods on PostBuild.

ryantrinkle:
It's conceivable, but currently the whole "notReady" infrastructure is pretty bound up in the details of how things are drawn.  However, it could potentially be turned into some kind of generic way of buffering an EventWriter.  The key would be to do this without a negative performance impact.

dalaing:
I was mostly thinking that the idea of "readiness" occurs in both PostBuild and NotReady, so it could save some one the mental burden of working out which typeclass to use if they were merged / reduce the number of concepts people have to take onboard.

Another possibility would be to have `type Ready t m = (PostBuild t m, NotReady t m)`, but I don't think that's all that useful.  It was mostly just a fleeting thought I had and figured I should mention :)

ryantrinkle:
Thanks!

src/Reflex/Workflow.hs
```diff
@@ -0,0 +1,37 @@
...
+  return $ fmap fst eResult
+
+workflowView :: forall t m a. (Reflex t, NotReady t m, Adjustable t m, MonadFix m, MonadHold t m, PostBuild t m) => Workflow t m a -> m (Event t a)
+workflowView w0 = do
+  rec eResult <- networkView . fmap unWorkflow =<< holdDyn w0 eReplace
+      eReplace <- fmap switch $ hold never $ fmap snd eResult
+  return $ fmap fst eResult
+
+mapWorkflow :: (Reflex t, Functor m) => (a -> b) -> Workflow t m a -> Workflow t m b
+mapWorkflow f (Workflow x) = Workflow (fmap (f *** fmap (mapWorkflow f)) x)
```
Ericson2314: Is it intentional that either of these are cannonized in a functor instance?

ryantrinkle:
I don't think so, but Call would know

ryantrinkle:
*Cale

PR #139, opened 2017-11-03
Relaxed haskell-src-exts upper bound for GHC 8.2.1

schell:
The upper bound of `haskell-src-exts` prevented stack from resolving deps on nightly. This bumps the upper bound to include `1.19`, which seems to work.


ryantrinkle:
Thanks!

PR #142, opened 2017-11-14
Add Monoid and Semigroup instances

### LINKED ISSUE #141: Monoid/Semigroup instances for PostBuildT
where `<>` == `>>`. This is handy in same cases for abstractions and simply because the precedence of these operators is different.

3noch:
Fixes #141

### DIFF
--- a/src/Reflex/PerformEvent/Base.hs
+++ b/src/Reflex/PerformEvent/Base.hs
@@ -43,6 +43,7 @@ import Data.Dependent.Sum
 import Data.FastMutableIntMap
 import Data.IntMap.Strict (IntMap)
 import qualified Data.IntMap.Strict as IntMap
+import qualified Data.Semigroup as S
 
 -- | A function that fires events for the given 'EventTrigger's and then runs
 -- any followup actions provided via 'PerformEvent'.  The given 'ReadPhase'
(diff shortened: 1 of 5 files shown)


3noch:
Of note: https://mail.haskell.org/pipermail/haskell-cafe/2017-March/126605.html

3noch:
Using `Alternative` for any of the monads in Reflex seems highly dubious. However, using `Monoid (m a)` seems seems possible.

ryantrinkle:
Thanks!

Ericson2314:
The mailing list post there pointed out there's multiple instances and neither is necessarily better, hmm.

ryantrinkle:
My recollection was that that was more or less resolved

3noch:
Some more argumentation was made in November regarding ReaderT: https://mail.haskell.org/pipermail/libraries/2017-November/028312.html

3noch:
And `IO` has this instance: https://hackage.haskell.org/package/base-4.10.1.0/docs/Prelude.html#t:Monoid

ryantrinkle:
Actually, looking at this code, there *is* one more thing we should change: we should also lift mconcat, sconcat, and stimes, since the underlying implementations may be more efficient.

3noch:
Good call. I can add that this weekend.

ryantrinkle:
Sounds good :)

PR #147, opened 2017-12-22
Add FunctorMaybe instance to Maybe

alexfmpe:


src/Reflex/FunctorMaybe.hs
```diff
@@ -24,6 +24,10 @@ class FunctorMaybe f where
   -- | Combined mapping and filtering function.
   fmapMaybe :: (a -> Maybe b) -> f a -> f b
 
+-- | @fmapMaybe = flip (>>=)
+instance FunctorMaybe Maybe where
+  fmapMaybe = flip (>>=)
```
3noch: `(=<<)`?

ryantrinkle: Good idea :)

alexfmpe: Derp. Fixed.

ryantrinkle:
Thanks!

Issue #149, opened 2017-12-29
Overly-restrictive depends


ag-eitilt:
I get that most of the development is happening on 0.5, but until that is released, 0.4 is still the "current" version on Hackage, and should be maintained. In particular, the upper limits on some of the build-depends no longer accept the newest versions, and so are not compatible with GHC 8.2.1. Fixing that does require slightly more than updating the .cabal, but of the ones that don't, I was able to build (and have the test suite pass, though the latter doesn't seem exhaustive) using:

- base-4.9.1.0 (my system's not set up for 4.10)
- dependent-sum-0.4
- these-0.7.4
- template-haskell-2.11.1.0 (see following note)
- syb-0.4.4

Unfortunately, I wasn't fully able to bring things up to date as I had some trouble getting template-haskell >= 2.12 (required for ghc-8.2.1) together, and the version of haskell-src-meta that would support that (and the associated haskell-src-exts) do introduce build errors. Anyone with a better knowledge of the codebase and a less restictive package system able to estimate the work required?

ag-eitilt:
Working on a patch for this myself, but the 0.4 release is just marked with a tag rather than a full branch -- 0.2 and 0.5 are the only ones that have the latter. Should I just base my pull request on the closest preceeding branch, or would it be possible to create a new branch there in the main repo?

ryantrinkle:
Just created branch named "0.4".  Thanks for working on this!

ag-eitilt:
No problem! Still have one more thing to check when I get home tonight, but the patch is nearly ready as far as I can tell.

ag-eitilt:
Well, the problem I had was using a rather... interesting build environment. That PR should be ready now.

ag-eitilt:
Patch has been merged and the new version uploaded to Hackage.

ag-eitilt:
Looks like I forgot to bump the upper bound of base past the version I have, to the correct one for ghc-8.2.1; would it be possible to file a revision on Hackage to fix that to be "<4.11"? Sorry!

dfordivam:
This should not be required as the 0.5 release is on its way

PR #150, opened 2017-12-31
Patch 0.4.0.1

ag-eitilt:
Primarily intended to support more recent versions of hackage dependencies as described in #149 , but also incorporating the fixes to the benchmarks introduced a bit further along the main branch.

Unfortunately, I was not able to test `template-haskell-2.12:0.0` which is required for full compatibility with `ghc-8.2.1`, as `th-lift-instances` (required indirectly by `haskell-src-meta`) is unable to build with that version. I would appreciate someone checking my work there, though, to be sure I'm not just missing an easy fix to that issue that doesn't involve modifying that code as well.

Finally, I apologise for the overly-large commit list; the branch is based on 77b2a505 (which seems to be the actual code on hackage), but I used a merge to get a small number of lines along with the proper blame annotations, and didn't realize it would pull in the entire history rather than just the few involved commits.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,17 +3,17 @@
 
 Reflex is an fully-deterministic, higher-order Functional Reactive Programming (FRP) interface and an engine that efficiently implements that interface.
 
-[Reflex-DOM](https://github.com/ryantrinkle/reflex-dom) is a framework built on Reflex that facilitates the development of web pages, including highly-interactive single-page apps.
+[Reflex-DOM](https://github.com/reflex-frp/reflex-dom) is a framework built on Reflex that facilitates the development of web pages, including highly-interactive single-page apps.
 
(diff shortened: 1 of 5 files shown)


ag-eitilt:
Failing build mentioned above is being tracked in bennofs/th-lift-instances#10

ag-eitilt:
We have determined that that issue is caused by my installing template-haskell separate from ghc, which I can only do because I have an old cabal. For those with a sane environment, it builds just fine.

ag-eitilt:
Looks like the CI failure is due to an error with the description "could not create session key: disk quota exceeded". Is that something I need to worry about, or is it just an issue with Travis?

ryantrinkle:
@ag-eitilt Don't worry about travis :) We're moving to CircleCI (reflex-platform is already moved), because we've had a lot of problems with it.

PR #155, opened 2018-01-09
Consistent naming for type variables

alexfmpe:


Quickref.md
```diff
@@ -47,7 +47,7 @@ Since MonadHold depends on MonadSample, any [S] function also runs in [H] contex
 [ ]   attachWithMaybe            :: (a -> b -> Maybe c) -> Behavior a -> Event b -> Event c
 [ ]   attachPromptlyDynWithMaybe :: (a -> b -> Maybe c) ->  Dynamic a -> Event b -> Event c
 [ ]   <@>                        ::                 Behavior (a -> b) -> Event a -> Event b
-[ ]   <@                         ::                        Behavior b -> Event a -> Event b
+[ ]   <@                         ::                        Behavior a -> Event b -> Event a
```
alexfmpe: Is this the same as `tag` ?

ryantrinkle: @alexfmpe Yes, I believe it is.  That should probably be mentioned in the docs.

ryantrinkle:
Thanks!

Quickref.md
alexfmpe: How would that best be presented? 
- Inline comment?
- Part of a list of simple identities? (e.g flips, specializations)

alexfmpe: Oh, I guess you meant it should *already* be in the code documentation. Indeed it is:
  https://github.com/reflex-frp/reflex/blob/develop/src/Reflex/Class.hs#L1137

ryantrinkle: Well, it might be nice to have something that makes it clearer here, too.  But yes, those docs are also super important.

PR #158, opened 2018-01-17
GHC 8.4 fixes

deepfire:
Tested on GHC 8.4.1-alpha1 (with jailbreaking applied to `reflex`).

### DIFF
--- a/reflex.cabal
+++ b/reflex.cabal
@@ -140,10 +140,10 @@ library
   if flag(use-template-haskell)
     cpp-options: -DUSE_TEMPLATE_HASKELL
     build-depends:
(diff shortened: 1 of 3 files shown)


ElvishJerricco:
For all the Semigroup things, I recommend following [the `-Wcompat` suggestion for this](https://prime.haskell.org/wiki/Libraries/Proposals/SemigroupMonoid#Writingcompatiblecode). Not using CPP is really nice.

matthewbauer:
Did anyone ever figure out what the haddock error is coming from?

ali-abrar:
Looks like similar changes already made it in and we're ghc 84 compatible. Thanks!

PR #163, opened 2018-01-29
Steer people away from `switchPromptly*`

Ericson2314:
Create non-prompt counterparts, even though these helper functions are dramatically simpler, so that no one mistakenly thinks we are endorsing the prompt versions over their non-prompt equivalents. This is important because the "prompt" semantics are rarely needed, yet are much more difficult to use
correctly.

Also, rename `switchPromptly` and `switchPromptOnly` to `switchHold*`. This helpers show how they are far more related to each other (having identical types) than they are to plain `switch`.

One small downside of this change is that it obscures relationship between `switchPromptlyDyn` and `switchHoldDyn`. This is deemd worth it as the use of the `current` behavior to avoid holding really does make a semantic difference: Old inner events prior to the first outer event firing will never be lost.

### DIFF
--- a/Quickref.md
+++ b/Quickref.md
@@ -159,7 +159,7 @@ For Events, the returned Event fires whenever the latest Event supplied by the w
 [ ]   switch            ::                  Behavior (Event a)  ->    Event a
 
 -- Flatten Dyanmic-of-Event to Event.  New Event is used immediately.
-[ ]   switchPromptlyDyn ::                   Dynamic (Event a)  ->    Event a
+[ ]   switchDyn         ::                   Dynamic (Event a)  ->    Event a
 
 -- Flatten Event-of-Event to Event that fires when both wrapper AND new Event fire.
 [ ]   coincidence       ::                     Event (Event a)  ->    Event a
@@ -182,7 +182,7 @@ For Events, the returned Event fires whenever the latest Event supplied by the w
 
 -- Similar to above, for Events.  Created Event initially tracks the first argument.
 -- At switchover, the output Event immediately tracks the new Event.
-[H]   switchPromptly    ::       Event a ->    Event (Event a)  -> m (Event a)
+[H]   switchHold        ::       Event a ->    Event (Event a)  -> m (Event a)
 ```
 
 ## Typeclasses to introspect and modify an FRP network.
--- a/src/Reflex/Class.hs
+++ b/src/Reflex/Class.hs
@@ -62,9 +62,10 @@ module Reflex.Class
   , EitherTag (..)
   , eitherToDSum
   , dsumToEither
-    -- ** Switching between 'Event's
-  , switchPromptly
-  , switchPromptOnly
+    -- ** Collapsing 'Event . Event'
+  , switchHold
+  , switchHoldPromptly
+  , switchHoldPromptOnly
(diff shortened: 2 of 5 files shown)


3noch:
Yes! Thank you. This has been a long standing problem. I created `switchTardy` in my code bases for this reason.

Quickref.md
```diff
@@ -182,7 +182,7 @@ For Events, the returned Event fires whenever the latest Event supplied by the w
 
 -- Similar to above, for Events.  Created Event initially tracks the first argument.
 -- At switchover, the output Event immediately tracks the new Event.
-[H]   switchPromptly    ::       Event a ->    Event (Event a)  -> m (Event a)
+[H]   switchHoldNever   ::       Event a ->    Event (Event a)  -> m (Event a)
```
ryantrinkle: This doesn't look right - shouldn't `switchHoldNever :: Event (Event a) -> m (Event a)` ?

src/Reflex/Class.hs
```diff
@@ -62,9 +62,9 @@ module Reflex.Class
   , EitherTag (..)
   , eitherToDSum
   , dsumToEither
-    -- ** Switching between 'Event's
-  , switchPromptly
-  , switchPromptOnly
+    -- ** Collapsing 'Event . Event'
+  , switchHoldNeverPromptly
+  , switchHoldNeverPromptOnly
```
ryantrinkle: The "*Only" variant should be the only one of these that exists, or perhaps we should immediately deprecate `switchHoldNeverPromptly`.  The original implementation of `switchPromptly` was a mistake that should not be propagated further.

src/Reflex/Dynamic.hs
```diff
@@ -181,10 +182,26 @@ count e = holdDyn 0 =<< zipListWithEvent const (iterate (+1) 1) e
 toggle :: (Reflex t, MonadHold t m, MonadFix m) => Bool -> Event t a -> m (Dynamic t Bool)
 toggle = foldDyn (const not)
 
+-- | Switches to the new 'Event' whenever it receives one.  Switching occurs
+-- *after* the inner 'Event' fires - so if the 'Dynamic' changes and both the
```
ryantrinkle: I think this should say `*after* the outer 'Event' fires`

src/Reflex/Dynamic.hs
```diff
@@ -181,10 +182,26 @@ count e = holdDyn 0 =<< zipListWithEvent const (iterate (+1) 1) e
...
+-- old and new inner Events fire simultaneously, the output will fire with the
+-- value of the *old* 'Event'.
+--
+-- Prefer this to 'switchPromptlyDyn' where possible. The lack of doing double
+-- work when the outer and (new) inner fires means this imposes fewer "timing
+-- requirements" and thus is far more easy to use without introducing fresh
+-- failure cases.
```
ryantrinkle: Might want to also mention that it has better performance than the prompt version.

src/Reflex/Class.hs
Ericson2314: Agreed to do this later separately; best not to mix up naming improvements with performance-impacting changes.

ryantrinkle:
Thanks!

PR #167, opened 2018-02-15
Added Semigroup instance for Dynamic.

eskimor:
Untested, but it compiles and it is Haskell - so it has to be correct.

### DIFF
--- a/src/Reflex/Class.hs
+++ b/src/Reflex/Class.hs
@@ -911,6 +911,14 @@ zipDynWith f da db =
         return $ Just $ f a b
   in unsafeBuildDynamic (f <$> sample (current da) <*> sample (current db)) ec
 
+instance (Reflex t, Semigroup a) => Semigroup (Dynamic t a) where
+  (<>) = zipDynWith (<>)
+#if MIN_VERSION_semigroups(0,17,0)
+  stimes n = fmap $ stimes n
+#else
+  times1p n = fmap $ times1p n
+#endif
+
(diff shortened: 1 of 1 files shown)


eskimor:
Hmmm - the travis errors don't really seem related to my changes.

alexfmpe:
@eskimor Yeah, Travis has been failing for a while here, pay it no heed. We're planning to switch to hydra sometime.

eskimor:
@alexfmpe hydra - neat :-)

src/Reflex/Class.hs
```diff
@@ -911,6 +911,14 @@ zipDynWith f da db =
         return $ Just $ f a b
   in unsafeBuildDynamic (f <$> sample (current da) <*> sample (current db)) ec
 
+instance (Reflex t, Semigroup a) => Semigroup (Dynamic t a) where
+  (<>) = zipDynWith (<>)
```
3noch: This is `liftA2` no?

ryantrinkle: I believe `liftA2` only recently became a method of `Applicative`.  For versions where it's as method, `liftA2` and `zipDynWith` are equivalent; for older versions, `zipDynWith` is more performant.

eskimor: should be, yes.

3noch: @ryantrinkle Makes sense. Thanks.

PR #174, opened 2018-03-06
Reflex.Time: Add throttle

ali-abrar:
@cgibbard was too lazy to make a PR so I'm doing it with his code.

### DIFF
--- a/src/Reflex/Time.hs
+++ b/src/Reflex/Time.hs
@@ -222,6 +222,39 @@ batchOccurrences t newValues = do
       delayed <- delay t toDelay
   return $ tag buffer delayed
 
+-- | Throttle an input event, ensuring that at least a given amount of time passes between occurrences of the output event. If the input event occurs too
+-- frequently, the output event occurs with the most recently seen input value after the given delay passes since the last occurrence of the output.
+-- If the output event has not occurred recently, occurrences of the input event will cause the output event to fire immediately.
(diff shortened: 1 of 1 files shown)


mulderr:
Just the other day I was thinking I need something exactly like that to throttle scroll events but couldn't immediatelly figure out how to do it. This is great!

ali-abrar:
@mulderr `debounce` or `batchOccurrences` might also do something like what you're looking for. The difference between `debounce` and `throttle` is that `debounce` does not allow the initial event through until the cooldown time has elapsed, whereas `throttle` lets the initial event through immediately and then applies the cooldown to subsequent events (until enough time has elapsed without events that we consider the next one an initiating event).

For example, if you had `debounce 1 =<< button "click me"`, and clicked on it rapidly, you'd receive no events until you stopped clicking for one second. With `throttle` you'd get the first click, and then another when you stop clicking for one second.

cgibbard:
(Sorry Ali for taking too long on this, haha!)

There's another difference between throttle and debounce, that throttle shouldn't wait until you stop clicking before letting another occurrence through, it should only wait for the timeout to elapse before firing again.

We could also generalise this code a little further with regard to what happens when multiple occurrences happen while throttled (i.e. debounce is to batchOccurrences what throttle is to something).

mulderr:
In that case, I think this *is* exactly what I need. I only care about the last occurrence, I don't want it too often because it may cause DOM manipulation but I also don't want to wait until it settles like `debounce` because things should update more or less continuously.

PR #178, opened 2018-03-13
Fix haddock parse error

mulderr:
Getting this error when building haddocks:

```
src/Reflex/Time.hs:233:13: error:
    parse error on input ‘-- * Immediate mode turns off, and the buffer is empty.’
```


ryantrinkle:
Thanks!

Issue #145, opened 2017-12-04
EventWriter and MonadDynamicWriter instances for EventWriterT and DynamicWriterT


3noch:
These two monads could have instances of each other but don't. The modules need to be teased apart into `Class`/`Base` (like other monads in Reflex) first in order to avoid an import cycle.

ryantrinkle:
Agreed

PR #190, opened 2018-05-01
Split out Base and Class for DynamicWriter/EventWriter

3noch:
This causes breakage downstream (in `reflex-dom-core`, for example). Should I add `Reflex.DynamicWriter` and `Reflex.EventWriter` modules that just re-export to minimize the effects of this change?

Ref #145


ryantrinkle:
Yes, I think you should make (deprecated) stub modules.  Otherwise, this looks great :)

3noch:
@ryantrinkle Deprecated modules have been added

ryantrinkle:
Awesome; thanks!

PR #191, opened 2018-05-06
Add convenience functions: 'ffor2' & 'ffor3'

alexfmpe:
Intended to replace idioms like `ffor (zipDyn a b) $ \(a,b) -> (...)`.

I would've liked to add some more arities but `Control.Applicative` only exports up to 3.
We should only add higher arities for `ffor` if we also do for `liftA`.


schell:
Hehe, glad to see I'm not the only one who likes this kind of thing.

PR #194, opened 2018-06-10
fix haddock failure in PatchDMapWithMove

Dridus:
in my last PR, I used the `-- ^` syntax with a GADT constructor and haddock doesn't like that.


ryantrinkle:
Thanks :)

Issue #199, opened 2018-07-17
Combinator property tests


Ericson2314:
It would be nice to test functions like `takeWhileE` against their list-based equivalents.

PR #200, opened 2018-07-17
Add `dropWhileE` and `takeDropWhileE` to compliment `takeWhileE`

Ericson2314:
- `takeWhileJustE`
 - `dropWhileE`
 - `takeDropWhileE`

`headE` is also added to `takeWhileE` as a performance optimization.

Issue #199 tracking being able to easily write tests for things like this.


Ericson2314:
Includes #202 to avoid silly conflict.

PR #203, opened 2018-07-18
Suppress nil patch events in QueryT as an optimization

kmicklas:
Note that this now requires the query type to have an `Eq` instance.


src/Reflex/Query/Base.hs
```diff
@@ -115,19 +108,19 @@ instance (Reflex t, MonadFix m, Group q, Additive q, Query q, MonadHold t m, Adj
           let p k bs = case Map.lookup k bs0 of
                 Nothing -> case bs of
                   -- If the update is to delete the state for a child that doesn't exist, the patch is mempty.
-                  Nothing -> return mempty
+                  Nothing -> return Nothing
                   -- If the update is to update the state for a child that doesn't exist, the patch is the sample of the new state.
-                  Just newBs -> sampleBs newBs
+                  Just newBs -> Just <$> sampleBs newBs
```
danharaj: Shouldn't this and the other Just's also be maskMaybe?

kmicklas: Done.

danharaj:
It was actually a question whether it was legit, not a request.

kmicklas:
I think it's pretty clearly legit semantically either way. I left it out initially because I figured those ones would be not `mempty` frequent enough to make the comparison not worth it. But this is totally baseless guessing.

ali-abrar:
Do we have a perf test case for this?

kmicklas:
@ali-abrar No, we don't.

Ericson2314:
1. https://github.com/haskell/containers/issues/608#issuecomment-473773583 the Non Empty changes are actually look quite performant so far, so we might eventually want a `maskMempty :: Something a b => a -> Maybe b`

2. That `Something` need not require all of `Eq`, For example, with anything map-like can use [`null` from Foldable](http://hackage.haskell.org/package/base-4.12.0.0/docs/Data-Foldable.html#v:null). Extending our patch class hierarchy to have an `isEmpty` (and `maskEmpty` if we wish to be able to have associated non-empty type per the above) can make this work.

ali-abrar:
@kmicklas I'd like to include this in the 0.6 hackage release. Can you add a note to ChangeLog.md since this is a breaking change?

PR #206, opened 2018-07-18
 Fix Adjustable for EventWriterT with simultaneous replace and tell events

kmicklas:
Previously if the widget being replaced has a `tellEvent` which fires at the same time as the resulting ('replace finished') event returned from `runWithReplace`, the moribund widget's `tellEvent` will still count, potentially even at the same time as a new widget's `tellEvent`.

### DIFF
--- a/test/EventWriterT.hs
+++ b/test/EventWriterT.hs
@@ -1,9 +1,11 @@
 {-# LANGUAGE FlexibleContexts #-}
-{-# LANGUAGE GADTs #-}
+{-# LANGUAGE RecursiveDo #-}
+{-# LANGUAGE ScopedTypeVariables #-}
 module Main where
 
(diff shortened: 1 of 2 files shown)


test/EventWriterT.hs
```diff
@@ -43,3 +47,19 @@ testSimultaneous pulse = do
   forM_ [1,3..9] $ \i -> runWithReplace (tellEvent ([i] <$ e0)) $ ffor e1 $ \_ -> tellEvent ([i+1] <$ e0)
   return ()
 
+-- | Test that a widget telling and event which fires at the same time it has been replaced
+-- doesn't count along with the new widget.
+testMoribundTellEvent
```
ryantrinkle: Nice name :)

PR #209, opened 2018-07-19
Format the Collection module

ElvishJerricco:
This is one module that most reflex users should probably skim through. The Collection functions are extremely important in any nontrivial reflex app. It was pretty hard to read, so I just ran [brittany](https://github.com/lspitzner/brittany/) over it and used haskell-mode in emacs for the comments.


luigy:
Thanks!

PR #211, opened 2018-07-23
Dedup adjustable

Ericson2314:
Continues PR #121 from @dalaing. Ryan deemed the breakage acceptable. @ElvishJerricco I'll rebase this on top of your PR if that gets merged first. (And indeed it makes sense to do non-breaking first.)

### DIFF
--- a/reflex.cabal
+++ b/reflex.cabal
@@ -72,6 +72,7 @@ library
     Data.WeakBag,
     Reflex,
     Reflex.Class,
+    Reflex.Adjustable.Class,
     Reflex.Collection,
     Reflex.Dynamic,
     Reflex.Dynamic.Uniq,
--- a/src/Reflex.hs
+++ b/src/Reflex.hs
@@ -6,6 +6,7 @@ module Reflex
   ) where
 
 import Reflex.Class as X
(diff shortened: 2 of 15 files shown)


dalaing:
I still quite like being able to point people at Reflex.Collection for the collection management functions. It seems more obvious than Reflex.Adjustable.Class, and I'm not sure that the collection functions are as closely tied to Adjustable as the other functions that appear in Reflex.XYZ.Class.

Ericson2314:
OK removed the second commit so leaving `Reflex.Collection` as-is for now. @dalaing FWIW I was thinking that all the things in `Reflex.Collections` are more saliently also incremental adjustments than helpers that assist with working with collections, and also that `Reflex.Collection` was your own plan B too. In any event, it's better to deal with the dedup and `Reflex.Collection` collection as separate concerns, so good to get that first commit in :).

dalaing:
I don't have strong opinions either way, just a small preference for Reflex.Collection from a discoverability / teaching point of view.  Depending on how the imports all work out it may not be a big deal where they sit, especially since they're in the QuickRef document.

I really like the main thrust of the change though :)

PR #214, opened 2018-08-01
Bug in relating to holds of Event in push created by that event?

saulzar:
In the process of writing tests for `buildDynamic` with `PushM` I discovered what I thought was an issue with the changed code, but...

If I feed the test input of `[1,2,3,4]` each event occurrence should start a new hold, so the `Behavior` at the end should have the state `[[1,2,3,4], [2,3,4], [3, 4], [4]]` 

What I'm seeing is this though `[[1,1,2,3,4], [2, 2,3,4], [3, 3, 4], [4, 4]]`  where the `foldDyn` seems to be acted on by the event running the current frame. Correct me if I'm wrong about this behaviour!

The test:
```
testInnerHoldsE :: forall t m. (Reflex t, MonadHold t m, MonadFix m)
           => Event t Int -> m (Behavior t [[Int]])
testInnerHoldsE e =  do
  d <- foldDyn snoc [] $
    pushAlways (\a -> foldDyn snoc [a] e) e
   return $ current (join (fmap distributeListOverDynPure d))
    where snoc x xs = xs ++ [x]
```

Regarding the changes for the deriving for MonadHold/MonadSample I was getting errors - e.g.

```
src/Reflex/DynamicWriter/Base.hs:88:156: error:
    • No instance for (MonadHold t m)
        arising from the 'deriving' clause of a data type declaration
      Possible fix:
        use a standalone 'deriving instance' declaration,
          so you can specify the instance context yourself
    • When deriving the instance for (MonadHold
                                        t (DynamicWriterT t w m))

```

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -1,7 +1,17 @@
 # Revision history for reflex
 
+## Unreleased
+
+* Generalize merging functions:
+  `merge` to `mergeG`, 
+  `mergeIncremental` to `mergeIncrementalG`, 
+  `distributeDMapOverDynPure` to `distributeDMapOverDynPureG`,
+  `mergeIncrementalWithMove` to `mergeIncrementalWithMoveG`.  
+
 ## 0.6.2.0
 
+* Fix `holdDyn` so that it is lazy in its event argument  
+  These produce `DMap`s  whose values needn't be `Identity`.
 * Stop using the now-deprecated `*Tag` classes (e.g., `ShowTag`).
 * Fix `holdDyn` so that it is lazy in its event argument.
 
--- a/reflex.cabal
+++ b/reflex.cabal
@@ -49,6 +49,7 @@ library
     data-default >= 0.5 && < 0.8,
     dependent-map >= 0.3 && < 0.4,
     exception-transformers == 0.4.*,
+    profunctors,
     lens >= 4.7 && < 5,
     monad-control >= 1.0.1 && < 1.1,
     monoidal-containers == 0.4.*,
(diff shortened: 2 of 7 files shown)


ryantrinkle:
This is the behavior I expected - that a hold would be updated by events firing during the frame in which they're created.  Would you mind adding a test that compares this against Reflex.Pure?  I'm pretty sure that's what it does as well.

https://github.com/reflex-frp/reflex/blob/b55c6079a9c5029ae84afa40ff3415d95411be9c/src/Reflex/Pure.hs#L180-L183

Saulzar:
Interesting. I guess it doesn't seem intuitive to me, any idea what the
reasoning for these semantics are?!

It just happens to be that the Dynamics which come from factorDyn are
firing on the same frame as they're created, so you get this (non intuitive
to me) situation where if I fold on the inner Dynamics the first one comes
out without the initial value event and subsequent ones do!

So my test with the buildDynamic was giving me
[[0, 1, 2, 3], [1, 1, 2, 3], [2, 2, 3], [3, 3]]

I realised afterwards that I'm sure I've seen tests in the CrossImpl which
did exactly this - I've added another one which is just the code above.

ryantrinkle:
I think that's a bug in `factorDyn`, actually.  I'm not 100% sure, but I think `factorDyn` should never produce simultaneous internal and external firings.

dfordivam:
I have added the test case from this PR with #263 
I am not sure whether we need to add the `Holds.hs` test case or the standalone deriving code.
As per the discussion above the behavior of the code is correct and consistent between pure and spider implementations.

@Saulzar how did you get the `[[0, 1, 2, 3], [1, 1, 2, 3], [2, 2, 3], [3, 3]]` ?

Saulzar:
I think that looks like the output of:

last <$> runAppB testInnerHolds (Just <$> [0..4])

See line 24 of tests/Holds.hs in the commit (though it's changed to [1..4]).

PR #215, opened 2018-08-02
buildDynamic as PushM

Saulzar:
I've done the bare minimum in terms of changing Spider, and it seems the queue for initialising holds already suffices to handle nested buildDynamics - the tests pass, one of them is focused on nesting, and they initialise fine in reverse order in a `do rec` block.

Perhaps needs some more careful examination of how this looping operates.

factorDyn modified to avoid sample as an example.

No change needed in Reflex.Pure because `PushM` and `PullM` are equivalent there!


test/semantics.hs
```diff
@@ -38,12 +38,10 @@ main = do
...
     allTests = concat
      [ makeGroup "micro" Micro.testCases
-     , makeGroup "subscribing (100,40)" (Focused.subscribing 100 40)
-     , makeGroup "firing 1000" (Focused.firing 1000)
-     , makeGroup "merge 100" (Focused.merging 100)
-     , makeGroup "fan 50" (Focused.fans 50)
+     -- , makeGroup "subscribing (100,40)" (Focused.subscribing 100 40)
+     -- , makeGroup "firing 1000" (Focused.firing 1000)
+     -- , makeGroup "merge 100" (Focused.merging 100)
+     -- , makeGroup "fan 50" (Focused.fans 50)
```
ryantrinkle: Is this intentional?

ryantrinkle:
This is really awesome; thanks @Saulzar!

Saulzar:
Nope! This is me not being able to figure out how to pass arguments into
the tests with new-test .. whoops.

ryantrinkle:
Thanks!

PR #218, opened 2018-08-15
Hide <> from Prelude

hololeap:
Since base-4.9.0, Prelude exports `<>` and this creates an ambiguous occurrence wtih `GhcPlugins.<>`


ryantrinkle:
Thanks!

PR #222, opened 2018-08-19
Build incremental

Saulzar:
I had this kicking around, so I figured I might as well add a couple of tests and submit it for completeness. It  seems useful for writing containers. One very simple thing it enables is a safe `mapIncremental` which is monadic, but avoids the `sample` which would otherwise be required.


lpsmith:
The breakage here is because this adds a new method to the Reflex class.  It also includes a default method for the implementation, but it does specify additional constraints,  and I'm not sure without further study what might not be covered by this default method;  however the breakage might be very minimal.

It also uncomments and fixes the type of another default method.

PR #223, opened 2018-08-23
Added `unzipE` as an alias for `splitE` and the corresponding `zipE`

marcotoniut:
`unzipE` breaks up an `Event` of tuples into a tuple of `Event`s that occur at the same time.
`zipE` creates an `Event` of a tuple of values supplied by two simultaneous `Event`s.

The reasoning for this is backed by the module `Data.Align`, which makes mention of the analogy that exists between the `align` family of functions for `These` and that one of `zip` for tuples.

Functions of the `split` family, on the other hand, and as a proposal for an eventual repurposing, could be used to signal whenever an `Event` matches a specific predicate, by taking the following signature:
```hs
split :: MonadHold t m => (a -> Bool) -> Event t a -> m (Event t (Event t a))
```
Or by making use of the `Splitter` data type

### DIFF
--- a/src/Reflex/Class.hs
+++ b/src/Reflex/Class.hs
@@ -53,6 +53,8 @@ module Reflex.Class
   , alignEventWithMaybe
     -- ** Breaking up 'Event's
   , splitE
+  , unzipE
+  , zipE
   , fanEither
   , fanThese
   , fanMap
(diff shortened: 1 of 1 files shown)


3noch:
We now have an instance for [`Zip`](https://hackage.haskell.org/package/semialign-1.1/docs/Data-Zip.html#v:zip) from `semialign`. But we should also add an instance for `Unzip`.

3noch:
Superseded by existing Zip instance and #420

3noch:
Sorry for the enormous delay!

3noch:
We now have Zip and Unzip instances!

Issue #208, opened 2018-07-18
Add default implementation of `traverseIntMapWithKeyWithAdjust`


kmicklas:
Although performance will be lost, `traverseIntMapWithKeyWithAdjust` can be implemented in terms of `traverseDMapWithKeyWithAdjust`. Providing a default will fix many currently broken instances such as `QueryT` (#204) and DOM builders in `reflex-dom`.

PR #224, opened 2018-08-24
Add default for traverseDMapWithKeyWithAdjust in terms of traverseDMapWithKeyWithAdjustWithMove

kmicklas:
This is related to #208. However, unlike that hypothetical default, this shouldn't have any serious performance concerns as far as I'm aware.


ryantrinkle:
Yes, this looks good.  And, it reveals a law!  We should demand that the output patches have the same structure as the input patches - typically this would mean keeping them in an abstract Traversable type.  We clearly need more structure than Traversable, but I wonder if this line of reasoning could help us figure out the bigger abstraction for this stuff.

PR #225, opened 2018-08-26
Generalize patchThatChangesAndSortsMapWith

mightybyte:
This is a trivial generalization of `patchThatChangesAndSortsMapWith` that doesn't do the sorting and allows the user to do it however they want.


ryantrinkle:
Thanks!

Issue #234, opened 2018-10-15
`tellEvent` has incorrect behavior when used inside a dynamic list


JBetz:
MVE. The value only increments after the first time you press the `+` button.

```haskell
counterW :: MonadWidget t m => m ()
counterW = mdo
  let update [_] cur = cur + 1
      update _ cur    = cur
  counterD <- foldDyn update 0 counterE
  (_, counterE) <- runEventWriterT $
    listWithKey (Map.singleton 0 <$> counterD) $ \_ counterListItemD -> do
      dynText $ pack . show <$> counterListItemD
      incE <- button "+"
      tellEvent $ pure <$> incE
  pure ()
```

_Originally posted by @JBetz in https://github.com/reflex-frp/reflex/issues/233#issuecomment-429658824_

3noch:
Hm, this is odd. If you instead build `counterD` from a separate "+" button above the `listWithKey` then the `tellEvent` fires many times *until* you change `counterD` at which point the `tellEvent`s stop having any effect. This interplay is hidden by your example because `counterD` is built recursively.

JBetz:
I don't understand. How can `tellEvent` be fired many times without changing `counterD`?

JBetz:
So `tellEvent` gets busted when the dynamic map updates, not merely from repeated usage. I see.

FWIW, I asked @ryantrinkle whether he knew of any changes from 0.3 to 0.5 that might effect this:

> the only change i know of is that, previously, when you had an Adjustable widget (dyn, widgetHold, etc.), during the frame when it is changing, both old *and* new events would propagate out of it. now, only old ones will.

Could that be relevant?

Saulzar:
Seems like it does not occur for dyn/runWithReplace, so it must be the list handling (traverseDMapWithKeyWithAdjust?)

JBetz:
@3noch, @Saulzar  I found the change that caused the breakage: https://github.com/reflex-frp/reflex/commit/ca50bb4cf1e85e1c980b3ddad0e1b0c483542349#diff-ca34969f303705bc74cc027f0eedb19f 

@kmicklas Can you preserve your fix while preventing this side effect?

kmicklas:
@JBetz Hmm, I'll take a look.

kmicklas:
@JBetz I've figured out the issue, just working on writing up a performant fix now.

PR #236, opened 2018-10-22
Add equivalent of testMoribunTellEvent for traverseDMapWithKeyWithAdjust

kmicklas:
This confirms that, although ca50bb4cf1e85e1c980b3ddad0e1b0c483542349 has caused #234, it fixes the original bug for `traverseDMapWithKeyWithAdjust` in addition to `runWithReplace`. (This will help ensure a fix to #234 does not cause a regression in that respect.)

### DIFF
--- a/reflex.cabal
+++ b/reflex.cabal
@@ -208,6 +208,7 @@ test-suite EventWriterT
   build-depends: base
                , containers
                , deepseq >= 1.3 && < 1.5
+               , dependent-map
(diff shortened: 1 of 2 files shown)


kmicklas:
Also note that this fails with ca50bb4cf1e85e1c980b3ddad0e1b0c483542349 reverted.

kmicklas:
Closing in favor of #239.

PR #238, opened 2018-10-24
Give [Event|Dynamic]Writer instances of each other

### LINKED ISSUE #145: EventWriter and MonadDynamicWriter instances for EventWriterT and DynamicWriterT

3noch:
Fixes #145


ryantrinkle:
Thanks!

PR #240, opened 2018-10-31
Add eventjoin

jappeace:
I basically always use this to collapse events. I thought it'd be convenient to have in reflex itself.


src/Reflex/Class.hs
```diff
@@ -817,6 +818,10 @@ fanThese e =
 fanMap :: (Reflex t, Ord k) => Event t (Map k a) -> EventSelector t (Const2 k a)
 fanMap = fan . fmap mapToDMap
 
+-- | switcHold never = eventjoin
```
endgame: switc**h**Hold

3noch:
I'm not sure how I feel about this. `switchHold never` isn't *that* much longer than `eventJoin`. It also lets you pick the initial event but `eventJoin` obviously doesn't. I don't think it helps to pretend that `Event` is a monad in the API when it's not. This is an easy enough function to write that anyone needing it can just include it in their project if they want it.

alexfmpe:
I think the name `eventJoin` is misleading here because `join = coincidence` - see https://github.com/reflex-frp/reflex/blob/74c1db3d977708bfe9ce743a4f76ea9ccd541e00/src/Reflex/Class.hs

jappeace:
@3noch the reason I initially made this was because I couldn't find the function that would collapse Event Event, coincidence does have that from a type signature perspective but didn't do what I wanted (it compiled, but didn't work). I've not encountered the case where I needed the additional power of specifying an initial event, but perhaps your situation is different.

jappeace:
@alexfmpe The type signature is quite different though, it's not going to be confused easily. But perhaps renaming it to something with switch would be good.

alexfmpe:
Aren't they both `Event t (Event t a) -> Event t a` ?

alexfmpe:
Ah I see, `switchHold never` has a `m` on the return

3noch:
Perhaps what we really need is to just make the docs of `coincidence` point you in the right direction...

jappeace:
@3noch sounds like a good idea

Ericson2314:
Very happy how this one turned out. May the docs not leave anyone else confused about this!

src/Reflex/Class.hs
```diff
@@ -286,7 +286,8 @@ class ( MonadHold t (PushM t)
   -- | Create an 'Event' that will occur whenever the currently-selected input
   -- 'Event' occurs
   switch :: Behavior t (Event t a) -> Event t a
-  -- | Create an 'Event' that will occur whenever the input event is occurring -- and its occurrence value, another 'Event', is also occurring
+  -- | Create an 'Event' that will occur whenever the input event is occurring -- and its occurrence value, another 'Event', is also occurring.
+  --   You maybe looking for 'switchHold' 'never' instead.
```
endgame: `s/maybe/may be`

endgame: I really like these type of haddocks, because they help a new user of a library to get the "lay of the land" - how to solve problems using the library, and develop a sense of "where everything is".

To that end, I suggest adding a little more detail as to why the user might be considering `switchHold never`. Perhaps something like this?

    Note: You may be looking for @'switchHold' 'never'@ instead, which gives you access to the most recent 'Event t a'.

endgame: Also: I don't remember how haddock handles spaces. You might want to add some newlines to the comment block so that your note starts on a new line. ([Anti-example: `networkHold`](http://hackage.haskell.org/package/reflex-0.6.3/docs/Reflex-Network.html#v:networkHold))

alexfmpe: I don't feel "most recent" accurately conveys the promptly/non-promptly concerns. I'd just make sure `switchHold` itself is documented properly and stick with the "you probably don't want this" note

endgame: I agree with your objection. If you can find some better words to capture the flavour, great, otherwise yeah that's fine.

PR #241, opened 2018-11-05
use upstream release of monoidal-containers

hSloan:


ryantrinkle:
@hSloan Does this work with all versions of monoidal-containers 0.4 and up?

luigy:
@ryantrinkle only `0.4.0.0` has been released to hackage in that range, but this looks safe to me. You think it should be using tighter/looser bounds?

ryantrinkle:
Thanks!

Issue #184, opened 2018-04-07
More Reflex.Time and related functions


lspitzner:
See https://gist.github.com/lspitzner/18e679006765b308e9dbc41b06e86bbb

- `gateGather`: an accumulating gate
- `maskCalmE`: behaves like throttle, but not controlled by time, but by input dynamic
- `fanBy`: `(eventsWhenFalse, eventWhenTrue) = fanBy pred inputE`
- `fanOn`: `(eventsWhenFalse, eventWhenTrue) = fanOn boolBeh inputE`
- `triggerDelay`: behaves like delay, but not controlled by time, but my a second input event
- `executeAsync1Calm`: perform an event asynchronously, but do not run more than one at a time
- `executeAsync1CalmWith`: with custom forker function
- `gateThrottle`: like `throttle` but without any accumulation. better for debouncing bursts.

All names open for discussion.

Only `gateThrottle` actually uses time, so I doubt that `Reflex.Time` is the right module, but not sure where else these would belong.

lspitzner:
I have added `conditionalBuffer` because it was necessary to fix the semantics of `executeAsync1Calm`(`With`). Preventing logic loops when inputs get connected to outputs is tricky.. hope things are fixed now.

lspitzner:
To be clear, I am ready to turn this into a PR, presuming that there is interest in having these included and that we decide on which module to put these in.

PR #242, opened 2018-11-05
Fancy throttle

3noch:


lspitzner:
This reminds me of https://github.com/reflex-frp/reflex/issues/184 which has some similar constructs.

Also, at a glance I wonder if there is much use to using Semigroup if you are not forcing the intermediate value.

ElvishJerricco:
That *is* a fancy throttle! Nice!

lspitzner:
Does this behave as expected for a delay of 5s and a example input signal that fires once every second? If I am reading things correctly, the delayed signals are not filtered, meaning that after 5 seconds, the delayed signal fires every second too, which turns immediate mode back on at the same rate. So you get 1 output, 5s pause, 4 merged outputs, and then every (other?) second some output. Whether it is every second or every other probably is not deterministic due to continuous-time/fuzzy delays.

I'd expect proper 5s intervals of silence, as you'd get if only the non-immediate-mode-entering signals trigger, after a delay, the back-to-immediate-mode switch.

ali-abrar:
@3noch (maybe a question for James) do we have a test that would answer @lspitzner's question?

We should definitely figure out how this interacts/overlaps with #184 before merging. 

@cgibbard may also have some thoughts as the throttle code here seems to be an upgraded version of something he wrote (that somehow never made it to develop..) https://github.com/reflex-frp/reflex/blob/4bdf22876975be3cf53deb6471d746f843569d4d/src/Reflex/Time.hs#L243

3noch:
cc @xplat

xplat:
> Does this behave as expected for a delay of 5s and a example input signal that fires once every second? If I am reading things correctly, the delayed signals are not filtered, meaning that after 5 seconds, the delayed signal fires every second too, which turns immediate mode back on at the same rate. So you get 1 output, 5s pause, 4 merged outputs, and then every (other?) second some output. Whether it is every second or every other probably is not deterministic due to continuous-time/fuzzy delays.
> 
> I'd expect proper 5s intervals of silence, as you'd get if only the non-immediate-mode-entering signals trigger, after a delay, the back-to-immediate-mode switch.

I had exactly the same question after reading the code for `throttle`.  @cgibbard gave me the answer -- the delay is fired based on the *output* events, not the *input* events.  So the first output, 5s pause, 4 merged outputs, and then nothing until 5 seconds after the 4 merged outputs.

It's all twisty, but it works.

lspitzner:
Right, i should have noticed that. sorry. Looks fine to me!

(and regarding functionality overlap: I think it is worth having this function in addition to the #184 ones. Even if you could express fancy-throttle in terms of `gateGather`, as I suspect you can.)

sboosali:
should we add that clarification to the documentation?

ali-abrar:
Given that the question has come up a couple times, I think we should

luigy:
pinging @cgibbard since I believe you were somewhat involved in this?

src/Reflex/Time.hs
```diff
@@ -35,6 +35,7 @@ import Data.Time.Clock
 import Data.Typeable
 import GHC.Generics (Generic)
 import System.Random
+import Data.Semigroup
```
3noch: The imports were alphabetized prior to this change

tomsmalley:
`throttleBatchWithLag pure` causes a hang which leads to a stack overflow. Granted, the docs say that the "lag" function should construct an event, but we should at least give a stronger warning there.

PR #248, opened 2018-12-22
Update README.md

dfordivam:


dfordivam:
Shoud we add link to obelisk in addition to (instead of) reflex-platform?

README.md
```diff
@@ -1,15 +1,15 @@
...
 Reflex is an fully-deterministic, higher-order Functional Reactive Programming (FRP) interface and an engine that efficiently implements that interface.
 
 [Reflex-DOM](https://github.com/reflex-frp/reflex-dom) is a framework built on Reflex that facilitates the development of web pages, including highly-interactive single-page apps.
 
-Comprehensive documentation is still a work in progress, but a tutorial for Reflex and Reflex-DOM is available at https://github.com/reflex-frp/reflex-platform and an introductory talk given at the New York Haskell Meetup is available here: [Part 1](https://www.youtube.com/watch?v=mYvkcskJbc4) / [Part 2](https://www.youtube.com/watch?v=3qfc9XFVo2c).
-
 A summary of Reflex functions is available in the [quick reference](Quickref.md).
 
-### Additional resources
+Visit https://reflex-frp.org/ for more information, tutorials, documentation and examples.
```
ElvishJerricco: I think it would be a good idea to make sure this link is directly under some header. People have a tendency to ignore all but the beginning of a section, unless it's a list of precisely what they need. This link is likely to be skimmed over in this position, and it's a very important link.

ElvishJerricco:
As for obelisk, I think that's worth noting in reflex-platform, but I strongly believe that obelisk should not be considered a primary function of reflex-platform. The platform should remain entirely independent of obelisk, only noting it as a suggestion, not as something that all reflex-platform users should prefer.

README.md
dfordivam: Made this bold

ryantrinkle:
> I strongly believe that obelisk should not be considered a primary function of reflex-platform

@ElvishJerricco Yes, I completely agree

PR #250, opened 2018-12-31
add Reflex.Network documentation to Quickref.md

goolord:
I use these functions frequently, and they seem underdocumented


luigy:
Thanks!

alexfmpe:
Maybe we should clarify their relation with https://github.com/reflex-frp/reflex-dom/blob/develop/Quickref.md#dynamic-widgets ?

PR #251, opened 2019-01-02
Fix the typo behvior, choose between a and the.

philderbeast:


ryantrinkle:
Good catch!  I do think it should say "the behavior", though, not "a behavior".

luigy:
Thanks!

PR #257, opened 2019-01-18
Fixes for haddocks

dfordivam:
In the inlined code like `Dynamic t (Map k a)`, I have made it italic/emphasis. This is a bit unconventional, but increases the readability.
Its a surprise that haddock doesn"t allow inline code with backticks. The only thing it supports is monospaced text (inside @), but this appears almost similar to the plain text.

### DIFF
--- a/src/Data/FastWeakBag.hs
+++ b/src/Data/FastWeakBag.hs
@@ -42,8 +42,8 @@ import Data.IORef
 import System.Mem.Weak
 #endif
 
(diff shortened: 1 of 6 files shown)


dfordivam:
image

luigy:
https://hydra.reflex-frp.org/eval/114593

luigy:
Thanks! that's unfortunate that there is no way to inline code, but italics look ok to me

PR #263, opened 2019-01-24
Add testcase of nested foldDyn from #214

dfordivam:


ali-abrar:
Does this test case fail or pass? Was there an associated fix for #214?

Issue #186, opened 2018-04-12
distributeListOverDyn duplication


pikajude:
http://hexagoxel.de/haddock/reflex-0.5/html/Reflex-Class.html#v:distributeListOverDyn

http://hexagoxel.de/haddock/reflex-0.5/html/Reflex-Dynamic.html#v:distributeListOverDynPure

I don't think we need both of these

ali-abrar:
Fixed via #264

PR #264, opened 2019-01-24
Mark distributeListOverDynPure as deprecated, remove duplication.

dfordivam:
I used the code from distributeListOverDynPure as it is easier to understand.


dfordivam:
Fixes #186

PR #268, opened 2019-01-30
Drop some upper bounds for the benefit of GHC 8.8

deepfire:


gwils:
Would you consider relaxing them instead of removing them entirely?

ali-abrar:
I'd be happy to accept a patch setting the upper bounds appropriately (but not dropping them entirely). Please let me know if you plan on doing this. Thanks!

PR #271, opened 2019-02-09
Add 'FunctorMaybe Option' instance

alexfmpe:


src/Reflex/FunctorMaybe.hs
```diff
@@ -9,6 +13,7 @@ import qualified Data.IntMap as IntMap
 import Data.Map (Map)
 import qualified Data.Map as Map
 import Data.Maybe
+import Data.Semigroup (Option(..))
 
 --TODO: See if there's a better class in the standard libraries already
```
3noch: :joy: Kinda late now I think...

alexfmpe:
Don't merge this yet. I should probably put a `#if` in the import itself

ryantrinkle:
Thanks!

PR #273, opened 2019-02-14
Add PrimMonad for RequesterT; Export some not-so-safe functions for l…

ali-abrar:
…ow-level Requester implementations


ali-abrar:
Needed by https://github.com/reflex-frp/reflex-dom/pull/275

PR #274, opened 2019-02-14
Relax bounds on 'these', add explicit import to avoid clash

### LINKED ISSUE #270: Loosen `these` bounds to 0.7.7?
Now that reflex-0.5 is on Hackage, nixpkgs master has it in hackage-packages.nix. Unfortunately, the version of `these` in nixpkgs is 0.7.6.

@luigy Recently tightened the upper bound for `these` from `< 0.8` to `< 0.7.6`. I'm having trouble driving the reflex build systems to actually test the relaxation. Is 0.7.6 safe for reflex, and if so, can someone please push a metadata rev to hackage?

endgame:
Fixes #270.

This passes `cabal new-test`, but has not been tested against ghcjs. I do not know how to update reflex-platform either, but this seems like a pretty safe change.

### DIFF
--- a/reflex.cabal
+++ b/reflex.cabal
@@ -60,7 +60,7 @@ library
     semigroups >= 0.16 && < 0.19,
     stm >= 2.4 && < 2.6,
     syb >= 0.5 && < 0.8,
-    these >= 0.4 && < 0.7.6,
+    these >= 0.4 && < 0.7.7,
     time >= 1.4 && < 1.9,
     transformers >= 0.2,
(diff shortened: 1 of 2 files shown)


endgame:
I do not have merge privileges.

ryantrinkle:
Thanks!

endgame:
Thanks for merging, Ryan. Any plans on another hackage release? It would be cool to have reflex in nixpkgs natively ( see https://github.com/NixOS/nixpkgs/issues/55758 ).

PR #278, opened 2019-02-24
Allow joining 'Dynamic's through 'Traversable' or 'Distributable'

alexfmpe:


reflex.cabal
```diff
@@ -46,6 +46,7 @@ library
     containers >= 0.5 && < 0.7,
     data-default >= 0.5 && < 0.8,
     dependent-map >= 0.2.4 && < 0.3,
+    distributive,
```
alexfmpe: Should be safe since our lower bound of `semigroupoids` already depends on `distributive`

src/Reflex/Dynamic.hs
```diff
@@ -232,12 +235,21 @@ distributeListOverDynPure =
...
 
---TODO: Generalize this to functors other than Maps
 -- | Combine a 'Dynamic' of a 'Map' of 'Dynamic's into a 'Dynamic'
 -- with the current values of the 'Dynamic's in a map.
 joinDynThroughMap :: forall t k a. (Reflex t, Ord k) => Dynamic t (Map k (Dynamic t a)) -> Dynamic t (Map k a)
 joinDynThroughMap = joinDyn . fmap distributeMapOverDynPure
 
+-- | Combine a 'Dynamic' of a 'Traversable' of 'Dynamic's into a 'Dynamic'
+-- with the current values of the 'Dynamic's in the functor.
+joinDynThroughTraversable :: (Reflex t, Traversable f) => Dynamic t (f (Dynamic t a)) -> Dynamic t (f a)
```
alexfmpe: See https://github.com/alexfmpe/reflex/pull/1

src/Reflex/Dynamic.hs
```diff
@@ -232,12 +235,21 @@ distributeListOverDynPure =
...
+joinDynThroughTraversable = join . fmap sequence
+
+-- | Combine a 'Dynamic' of a 'Distributive' of 'Dynamic's into a functor of 'Dynamic's.
+-- The current value at each hole in the resulting functor will be the current value of that hole in the current original functor
+joinDynThroughDistributive :: (Reflex t, Distributive f) => Dynamic t (f (Dynamic t a)) -> f (Dynamic t a)
+joinDynThroughDistributive = fmap join . distribute
```
alexfmpe: I believe `distribute` will make all the child `updated` events trigger whenever the parent one does. 
Is this what we want? Do we need `holdUniqDyn` in there?

src/Reflex/Dynamic.hs
ryantrinkle: This looks like it will have bad performance characteristics.  `joinDynThroughMap` uses a single `merge` and ensures that changes are applied about as efficiently as possible to the output.  We've gotta do something like that here, otherwise we can't recommend that people do this generally.

src/Reflex/Dynamic.hs
```diff
@@ -243,7 +245,21 @@ joinDynThroughMap = joinDyn . fmap distributeMapOverDynPure
...
+
+    -- Requires the list to have at least as many elements as the traversable
+    unsafeReifyContents :: Traversable t => t a -> [b] -> t b
+    unsafeReifyContents = evalState . sequence . (f <$)
+      where
+       f = do
+         (b:bs') <- get
+         put bs'
+         return b
+
```
ryantrinkle: Let's make the failure cases explicit, with `error` calls: 1) if `get` returns an empty list, or 2) if anything is left over after the state monad is done running.

alexfmpe: Would it be better to instead error out at the start if the traversable and list do not have the same length?

src/Reflex/Dynamic.hs
ElvishJerricco: @ryantrinkle I don't think it gets better than `joinDynThroughTraversable = (>>= sequenceA)` with only the `Traversable` constraint. The problem is the cost of small updates to the potentially large container. But there's no way to do that better than O(n) using only `Traversable`. Even if you use something like `distributeListOverDynPure`, you still have to reify the new list into the original container type, which will be O(n).

src/Reflex/Dynamic.hs
ApolloUnicorn: @Ericson2314 , @ryantrinkle any thoughts on the question above?

src/Reflex/Dynamic.hs
xplat: Can't do better without extra constraints.  In fact, not sure we should have this because almost 100% of the time it will be better to have the Dynamic be inside the f in the first place, *especially* for Distributive where there's nothing outside the 'holes' that can change ...

src/Reflex/Dynamic.hs
xplat: `sequenceA` ought to accumulate efficiently anyway unless the datatype is inherently biased against that (e.g. lists)

3noch:
Closing this for now until a more obviously good option becomes apparent. We don't want to provide generalized combinators that we can't actually recommend people use.

PR #279, opened 2019-02-24
Add hlint file

alexfmpe:


ryantrinkle:
This looks partially redundant with https://github.com/reflex-frp/reflex/blob/2e20b188761cdf13909b7edfc6788241682c0ee4/test/hlint.hs .  Is there any way we can get the best of both?

alexfmpe:
I've run into that file and quickly browsed its hackage, but couldn't figure out how to add new hints to it. The good news is that the test suite hlint *does use* the .yaml file

ryantrinkle:
I see - you're actually adding new rules.

Now that I've looked at it more closely, I don't quite see why these rules would be good.  Can you explain the advantage of these rewrites?

ryantrinkle:
Although I think the goal and approach (i.e. adding hlint rules) here is good, I find these specific rewrites less legible than the code we have now.  I'm going to close this out, but we should keep this approach in mind in the future.

alexfmpe:
I added the `preview` ones because I see utils for them being defined quite regularly, so I saw there was a *demand* (doesn't imply this was the proper *supply*).

The `mfilter` one is definitely less popular - the advantage in my book is that it signals the intent of syncing the payload and the target of the predicate. You can't accidentally change one (to something of the same type) and forget the other one.
I'd argue the fact that it's less used isn't a problem because I see one of the advantages of hlint rules as being exactly teaching you equivalences for reuse.

alexfmpe:
These hints were just merged into hlint itself, into rule groups that are disabled by default. That's good enough for me, and keeps this repo on the standard track.
On another note, I think we ought to make a suite of reflex-specific hints to help with discoverability.

PR #286, opened 2019-03-17
Deprecate FunctorMaybe in favor of Filterable

ali-abrar:
Deprecates FunctorMaybe.

The class is still re-exported by Reflex.Class and hence by Reflex, but the method fmapMaybe is not automatically re-exported (in favor of exporting a new fmapMaybe alias to mapMaybe from Filterable). This isn't ideal, as it'll still cause some breakage where instances were defined (I think).

### DIFF
--- a/reflex.cabal
+++ b/reflex.cabal
@@ -1,5 +1,5 @@
 Name: reflex
-Version: 0.5.0.1
+Version: 0.6
 Synopsis: Higher-order Functional Reactive Programming
 Description: Reflex is a high-performance, deterministic, higher-order Functional Reactive Programming system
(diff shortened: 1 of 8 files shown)


ali-abrar:
`reflex-dom-core` compiles against this with no errors or new warnings.

reflex.cabal
```diff
@@ -1,5 +1,5 @@
 Name: reflex
-Version: 0.5.0.1
```
ryantrinkle: I think this needs to be 0.6, since deprecating `FunctorMaybe` is a breaking change.

PR #295, opened 2019-03-28
Some more haddocks

ali-abrar:


src/Reflex/Adjustable/Class.hs
```diff
@@ -11,6 +11,13 @@
 #ifdef USE_REFLEX_OPTIMIZER
 {-# OPTIONS_GHC -fplugin=Reflex.Optimizer #-}
 #endif
+-- |
+-- Module:
+--   Reflex.Adjustable.Class
+-- Description:
+--   A class for actions that can be "adjusted" over time based on some 'Event'
+--   such that, when observed after the firing of any such 'Event', the result
+--   is as though the action was originally run with the 'Event's value.
```
3noch: This is kinda confusing. It seems like it's saying "After applying Event to an Adjustable it's as if the Event was applied..." which isn't really useful.

ali-abrar:
I agree it could be clearer. Suggestions welcome.

src/Reflex/Adjustable/Class.hs
3noch: The description on already on `Adjustable` seems pretty good: https://github.com/reflex-frp/reflex/blob/62b3f1bf202b74b8bf4c3aa77afb4338b398c70e/src/Reflex/Adjustable/Class.hs#L45-L50

PR #297, opened 2019-03-29
Use reverse chronological order in ChangeLog

3noch:
A preference of mine so that I see the most recent changes first when I open the file in GitHub, etc. When in a termal, I'd use `less` anyway. Feel free to close if you disagree.


3noch:
This also seems to be a convention. Most changelogs I read are in this order.

ali-abrar:
Thanks!

PR #298, opened 2019-03-31
WIP: Add the SwitchHold class for things that can switch on Event firings

ali-abrar:


ali-abrar:
This could probably help clean up some code in other places, but before doing any of that, I thought it might be good to let others have a look and comment. 

CC @cgibbard @ryantrinkle

src/Reflex/Class.hs
```diff
@@ -299,6 +302,32 @@ class ( MonadHold t (PushM t)
...
+-- cyclic dependency / metastability failures. It's also more performant. Use
+-- this rather than 'switchHoldPromptly' and 'switchHoldPromptOnly' unless you
+-- are absolutely sure you need to act on the new event in the coincidental
+-- case.
+instance Reflex t => SwitchHold t (Event t a) where
+  switchHold ea0 eea = switch <$> hold ea0 eea
+
+-- | Switches to the new 'Dynamic' immediately (i.e., promptly) on firing
+instance Reflex t => SwitchHold t (Dynamic t a) where
+  switchHold a e = join <$> holdDyn a e
```
Ericson2314: I thought join for dynamics was prompt?

Ericson2314: Something with regular `hold` and `unsafeBuildDynamic` probably has the correct semantics and is the most efficient.

ryantrinkle: Please don't use `unsafeBuildDynamic` unless you can also prove that the result obeys the Dynamic law, i.e. that the Behavior changes when, and only when, the Event fires.

Ericson2314:
Not to squash the side benefit of steering away people from promptly, but we could make a class for that too just to get rid of the redundant names there too? Hopefully the benefit of having fewer names to choose between would "give people more time to think" and not just end up with people being overly prompt again.

src/Reflex/Class.hs
Ericson2314: Of course! So what I concretely came up with is:
```haskell
import Control.Monad
import Reflex

switchHoldDynamic
  :: (Reflex t, MonadHold t m)
  => Dynamic t a
  -> Event t (Dynamic t a)
  -> m (Dynamic t a)
switchHoldDynamic d ed = switchDynamic <$> hold d ed

-- `switchDyn` and this can duke it out re the naming convention.
switchDynamic
  :: Reflex t
  => Behavior t (Dynamic t a)
  -> Dynamic t a
switchDynamic bd = unsafeBuildDynamic (sample b) e
  where
    b = join $ current <$> bd
    e = switch $ updated <$> bd
```
This leads me to think:
1. Every "unprompt" `switchHold` begins with a plain `hold`. Throwing away the outer event thereafter and just keeping the behavior is crucial to avoiding any promptness / strictness.
2.  If all the interesting logic is in the switch part, perhaps we also want to make `Switch` a class, and today's `switch` rename to `switchE` or something.
3. Single `join` is "`switchB`", `switchDynamic` is actually `unsafeBehaviorEventPairToDyn . bimap switch switch . dynToBehaviorEventPair`.

I won't claim to prove this correct, I think per the above the single `Behavior t (Dynamic t a)` means promptness is impossible, and switching "preserves syncronization" so that switching the underling behavior and event of the inner dynamic maintains the invariants. Put another way, if switching didn't preserve that, I would question the value of `Switch` or `SwitchHold` classes.

ryantrinkle: `switchDynamic` produces an invalid `Dynamic`.

```haskell
-- | If the input event ever fires, the output Dynamic will misbehave.
makeBroken :: (Reflex t, MonadHold t m) => Event t () -> m (Dynamic t Int)
makeBroken changeOuter = do
  outer :: Behavior t (Dynamic t Int) <- hold (pure 0) $ pure 1 <$ changeOuter
  d <- switchDynamic outer
  return d
```

In the invocation of `switchDynamic`, `e` will never fire, because both input `Dynamic`s (`pure 0` and `pure 1`) are constant.  As a result, `updated d` will never fire.  Nevertheless, the value of `current d` will change.

ali-abrar:
I marked this WIP for now because it requires more thought about what the semantics/laws of this class really ought to be.

Saulzar:
I use a type class in my applications named exactly the same way.

 (I also have a class for Switch and SwitchPromptly). I find it quite
useful because I can also have instances for tuples so for example it also
allows switching several unrelated things (e.g. a mix of events, behaviours
and dynamics).

I think there should be two classes for SwitchHold in the same way,
*SwitchHold* (with non prompt switching) *and* *SwitchHoldPromptly* (prompt
switching)

FWIW - this is my instance for Dynamic. It implements _non prompt_
switching. It was written a while ago so I haven't looked too closely just
now to see what it does!
instance Reflex t => SwitchHold t (Dynamic t a) where
switchHold d ed = do
let eb = current <$> ed

b <- switchHold (current d) eb
e <- switchHold (updated d) (updated <$> ed)
buildDynamic (sample b) (pushAlways sample eb <!> e)

ali-abrar:
@Saulzar I think a clean separation between the prompt and non-prompt versions makes sense.

ryantrinkle:
@ali-abrar Agreed.

@Saulzar When this version of `switchHold` switches to a `Dynamic` whose `updated` is firing at the moment of the switch, won't this result in the new value of the new `Dynamic` getting lost, and then the output `Dynamic` staying "stale" until the next time either that `Dynamic` changes or the input `Event` fires?

Saulzar:
Doesn't that get captured in 'e'? Specifically in 'updated  <$> ed'?

e <- switchHold (updated d) (updated <$> ed)

It's using buildDynamic (not unsafeBuildDynamic) so the output behaviour
will 'hold' that value, too.

It's been ages since I looked at it, so I may be wrong - but I do remember
spending quite a long time trying to fix this instance!

src/Reflex/Class.hs
Ericson2314: Sorry for my wrong-headed attempt. This makes me wonder if there is non-prompt `switchHoldDyn`. The behavior goes with the prompt join of dynamics. The event is suitably not strict/prompt, but doesn't seem to really respect the original dynamic.

Ericson2314:
@Saulzar btw your markdown is broken for me. I've seen GitHub do this before for certain users, I think it has something to do with responding by email.

Saulzar:
Actually I think you're right - it would lose that value? These things are
complex and need some thought to get right...

I think that's what this was meant to solve - but I think in the case the
Dynamic is firing at the frame of the switch this will sample 'too early'
pushAlways sample eb

ali-abrar:
I'm starting to feel like it isn't, in which case the usefulness of the prompt-only class (which can't include an instance for Behavior) or non-prompt-only class (which can't include an instance of Dynamic) is somewhat less clear. The original motivation for this was to be able to write instances for tuples and other more complicated datastructures, but these exclusions would vitiate much of that benefit.

Ericson2314:
I think one way to think about is that switching for events can be strict+prompt or non-strict+non-prompt, and switching for behaviors is just the odd non-strict+prompt. (It's sort of weird/wrong to think of strict/prompt without nested events, but Behavior `join` definitely matches Dynamic `join` and not some non-prompt construction.)

cgibbard:
I'm a bit frustrated about this. Here's the situation:

* A dynamic must fire its updated Event every time it changes.
* A dynamic must switch its current value to the value that it fired its updated Event event with.

These rules can't be changed, they're part of the definition of Dynamic. In addition, we have:

* At any moment that the Event t (Dynamic t a) is not firing, we want the result of our switching operation to behave identically to the last Dynamic t a that it fired with, i.e. the value of its current Behavior should be identical, and its updated Event should be firing if that Dynamic's updated event is firing.

These three bullet points taken in conjunction imply that the resulting Dynamic's updated Event must fire with the new Dynamic's value. If the new Dynamic is changing at the same moment as the input Event of Dynamics is firing, our output Dynamic's updated Event *must* fire with the same value that the new Dynamic's updated Event is firing with, otherwise its current value would be incorrect immediately thereafter -- it won't have changed to the new Dynamic's value.

We have no real means of infinitesimally delaying the change, since infinitesimal delays are a bad idea for various reasons.

So, apart from things which don't mean anything close enough to what we want, we're stuck with only a prompt switching operation for Dynamic. For Event, we have prompt and non-prompt switching operations, and for Behavior, we have only non-prompt.

The frustrating thing is that pragmatically, we typically want the operation which has the least promptness possible, and there would be significant value in having an interface whereby we could provide a means to switch records of FRP-based gadgets wherein we pick the non-prompt operations for Event and Behavior (well, the only possible thing for Behavior), and then take the prompt switching for Dynamic simply because it's the only thing available. Another pragmatic point is that we often know that the occurrences of the Event *cannot* be at times when the contained things are changing, because we're e.g. switching between views of a user interface, and the user will not be interacting with the widgets contained therein at the same moments they are clicking a tab to switch between them.

But that said, indeed, it would be a very confusing inconsistency when anyone did trip over it.

So the options are:

* Define a pragmatic "the least prompt operation available" class that would often be very useful and clean up a common pattern that comes up in our code a fair amount, but has some confusing semantics in some cases that, while I think it would rarely bite people, would bite pretty hard when it did.
* Define a couple of separate classes, one of which does prompt operations and only supports Event and Dynamic, but we wouldn't want people to actually use because they'd be switching their Events in a needlessly prompt fashion most of the time. The other would only support Event and Behavior and so wouldn't be very useful in reflex-dom. That latter non-prompt class might be useful in some non-reflex-dom projects, but until we write one and see how it works out, I'm not sure we want to put it in. The former, prompt class is just harmful to have around, and I don't support adding it. It looks like the right thing, but it is not.
* Do nothing.

It's annoying.

Ericson2314:
@cgibbard absolutely agree with all of that. I think events and dynamics nest poorly in general; I don't really think it works well to nest mixed "patch -> dynamic" and "dynamic -> patch" widgets (I forget the tentative names for those idioms).

So one further avenue I think is to explore stuff related to `factorEvent`, which, DSums aside, makes a `Event t (_, Event t _)`. It has no infra outside of `factorEvent` so it's definitely an huge unknown, but avoids some promptness/strictness issues by providing that pure initial value. At the time it came about I had very vague notions in my head about how this relates with `Adjustable` and the `Reflex.Collection` menagerie. That sounds like related problems to what this was supposed to eventually address.

Saulzar:
@cgibbard - Well summarised, it's a contradiction between the semantics of Dynamic and the desired semantics of the operation in question.

Yet it's also very useful - I use it in my program in several places! It's not an issue because the Dynamics involved are never firing on the first frame.. but could lead to behaviour very hard to understand!

cgibbard:
> @cgibbard - Well summarised, it's a contradiction between the semantics of Dynamic and the desired semantics of the operation in question.
> 
> Yet it's also very useful - I use it in my program in several places! It's not an issue because the Dynamics involved are never firing on the first frame.. but could lead to behaviour very hard to understand!

Your operation is just a little different from what we're thinking of -- at least if I understand it correctly.

What happens if you try to switch between a bunch of constDyns?

Saulzar:
That's fine - `pushAlways sample eb` causes an event which occurs with the
switching event. The only trouble is when the new dynamic is firing an
event at the same time, in which case that code is sampling from the old
value (of the new dynamic).

saulzar:
I think this is worth resurrecting, even if Dynamic is not included - switching products of Event/Behaviour together are quite useful. Including a second class switchHoldPromptly in addition would seem a little ugly but it's one option?

It also occurs to me that Adjustable for  BehaviorWriterT vs. DynamicWriterT may have the same dilemma..

PR #300, opened 2019-04-07
WIP: Add common instances for 'Workflow'

alexfmpe:
I was writing `Workflow $ pure (a, never)` a lot in some placeholder code. One thing lead to another and this is what came out of that rabbit hole.

I'll do ~haddocks~ changelog/quickref parts when it's clearer if and which instances we're keeping.

~The `Alt` instance in particular is extremely questionable. The output workflow is replaced whenever either input workflow replacement is triggered, so it's a "last-updated" kind of thing. I don't know how to judge whether this instance is appropriate, since I couldn't find any laws to use as criteria.~
`Alt` actually does list its laws, it's `Alternative` that doesn't seem to.

If anyone wants to mess with these a bit, just do `ob run` in the skeleton folder of https://github.com/alexfmpe/obelisk/tree/test-workflow-instances and visit `http://localhost:8000`

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 * Fix `holdDyn` so that it is lazy in its event argument
+* Relax constraint from `Applicative` to `Apply` in `ffor2` & `ffor3`. `Event`s and `Map`s can now use these utils.
 
 ## 0.6.1.0
 
--- a/Quickref.md
+++ b/Quickref.md
@@ -59,6 +59,9 @@ Since MonadHold depends on MonadSample, any [S] function also runs in [H] contex
 [ ]   mergeList  :: [Event a] -> Event (NonEmpty a)
 [ ]   merge      :: GCompare k => DMap k Event -> Event (DMap k Identity)
 [ ]   mergeMap   :: Ord k => Map k (Event a) -> Event (Map k a)
+[ ]   ffor2      :: Event a -> Event b ->            (a -> b -> c)      -> Event c
+[ ]   ffor3      :: Event a -> Event b -> Event c -> (a -> b -> c -> d) -> Event d
+
 
 -- Efficient one-to-many fanout
 [ ]   fanMap    :: Ord k      => Event (Map k a)         -> EventSelector (Const2 k a)
@@ -98,7 +101,7 @@ Since MonadHold depends on MonadSample, any [S] function also runs in [H] contex
 [ ]   <>   :: Monoid a => Behavior a ->        Behavior a -> Behavior a
 -- ... plus many more due to typeclass membership
 
--- Combine multiple behaviors via applicative instance
+-- Combine multiple behaviors via 'Data.Functor.Plus.Apply' instance
 [ ]   ffor2 :: Behavior a -> Behavior b ->               (a -> b -> c)      -> Behavior c
 [ ]   ffor3 :: Behavior a -> Behavior b -> Behavior c -> (a -> b -> c -> d) -> Behavior d
 
@@ -139,7 +142,7 @@ Since MonadHold depends on MonadSample, any [S] function also runs in [H] contex
 [ ]   >>=                       ::                  Dynamic a -> (a -> Dynamic b) -> Dynamic b
 [ ]   zipDynWith                :: (a -> b -> c) -> Dynamic a -> Dynamic b        -> Dynamic c
 
--- Combine multiple dynamics via applicative instance
+-- Combine multiple dynamics via 'Data.Functor.Plus.Apply' instance
 [ ]   ffor2 :: Dynamic a -> Dynamic b ->              (a -> b -> c)      -> Dynamic c
 [ ]   ffor3 :: Dynamic a -> Dynamic b -> Dynamic c -> (a -> b -> c -> d) -> Dynamic d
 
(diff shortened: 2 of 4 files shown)


alexfmpe:
Here's a very handwavy-ish attempt at reasoning about which laws hold.

### Terminology
A *descendant* workflow is one created by another workflow's event triggering.
*Payload* of a workflow refers to the value emitted in its initial instant, and the value emitted by its descendant workflows in their initial instants. Basically what the `workflow` function returns.

A workflow is said to be:
- *replaced*: if it's event is being triggered. Basically `workflowView` or `fmap updated . workflow`
- *firing*: if it's at its initial instant, be that due to replacing another workflow, or being a root workflow (not descendant of another)

Crucially, `parallelWorkflows` will always run both input workflows in parallel, in an associative way. This means we need only check the payloads themselves match the laws for Apply/Alternative/Semigroup. Also, given that when no input workflow is firing, then neither is the output workflow, I'll only consider the cases when at least one input workflow is firing (otherwise, the payloads are constant and equalities are necessarily preserved at least until the next firing).

### Functor
Derived by GHC

### Alt ([formulation](https://hackage.haskell.org/package/semigroupoids-5.3.2/docs/Data-Functor-Alt.html#t:Alt))
#### Associativity:
#### `(a <!> b) <!> c = a <!> (b <!> c)`
When only one input workflow is firing, its payload is used for the output workflow, and we ignore the others.
In case of simultaneous occurrences, the leftmost workflow will always provide the resulting workflow, regardless of parenthesization.

#### Left-distributivity:
#### `f <$> (a <!> b) = (f <$> a) <!> (f <$> b)`
When A and B are both firing we have `f <$> (a <!> b) = f <$> a = (f <$> a) <!> (f <$> b)`
When only one is firing, we simply ignore the other and simplify the equations in the same way

### Apply/Applicative ([formulation](https://wiki.haskell.org/Typeclassopedia#Alternative_formulation))
Let 
- `(**) = zipWorkflows`
- `unit = pure ()`

#### Left (resp. right) identity: `unit ** v ≅ v`
This seems exactly the same as `fmap (,) () v` (so we have the isomorphism). At least the difference can't be observed by `workflow` or `workflowView`, since the unit workflow never triggers a replacement.

#### Associativity: `u ** (v ** w) ≅ (u ** v) ** w`
Parenthesization does not affect the payload timings, only the values themselves, which are isomorphic tuples.

### Semigroup
As long as the underlying semigroup is associative, so is the one over workflows due to directly delegating via `Apply`


### Bind/Monad ([formulation](https://en.wikibooks.org/wiki/Haskell/Category_theory#The_monad_laws_and_their_importance))
Here, we're looking at associativity along a different axis - in case of simultaneous occurence, the outer workflow "wins" and is replaced with its descendant: the event of the inner workflow fires but gets dropped due to being second in `leftmost`.

#### 1. `join . fmap join = join . join`
The order in which we collapse layers is irrelevant:
- Resulting payload: the one of the innermost workflow
- Replacement timings: the outermost workflow always wins in case of conflict. Otherwise, whatever layer is firing operates as usual.

#### 2. `join . fmap return = join . return = id`
LHS: inject the constant workflow inside another one, then collapse
RHS: inject another workflow inside the constant one, then collapse
Collapsing the layers on both cases has the same effect, since the constant functor is never replaced, and the innermost initial payload will be the same in both cases.

#### 3. `return . f = fmap f . return`
Applying `f` before or after lifting into the constant workflow yields the same result.

#### 4. `join . fmap (fmap f) = fmap f . join`
This one states that applying `f` to the inner workflow and then collapsing is the same as collapsing and then applying `f` to the resulting (with one less layer) workflow. Since `fmap` cannot affect timings, and `join` cannot affect payloads, we can shuffle them around without interference.

Ericson2314:
LGTM, Can we get `SemiAlign` and `Align` instances too?

alexfmpe:
`Semialign` requires `these >= 0.8` and breaks backward compatibility by moving methods out of `Align`. Bumping `these` in reflex ought to be its own PR, I reckon.

As for `Align` (after or before 0.8), it requires `nil`, which can't be implemented for `Workflow` due to the initial payload. The same happens for `Alternative/empty`.

Similarly, `FunctorMaybe` doesn't work - what would `fmapMaybe (const Nothing)` yield? The payload must be of type `b`, not `Maybe b`. A hypothetical `Filterable` typeclass with just a `filter :: (a -> Bool) -> f a -> f a` could be instantiated, by restricting its effect to the replacing-event. Perhaps we could add a `filterWorkflow` that just filters transitions?

On another note, since `Workflow t m` is already wrapped in `m`, we can make it an instance of `FunctorWithIndex`, even though `zipListWithEvent` being monadic prevents the same for `Event t`. The index is the number of replacements/transitions made by that workflow.

alexfmpe:
I just found this: https://github.com/reflex-frp/reflex/commit/7ea831d64246638fbea174e3682cd2ab4cd2bd16

alexfmpe:
Turns out that this formulation for collapsing parallel/nested workflows is fundamentally wrong - it doesn't preserve inner state (e.g. form input) because the entire composite workflow is replaced every time. I didn't notice it immediately because the ones I was testing with had no inner state.
It seems I managed to do this properly with a slightly different structure - will open another PR in the coming weeks.

PR #305, opened 2019-05-02
runHost function

ElvishJerricco:
Adding a function for creating new hosts with PerformEvent and PostBuild easily. Currently having some trouble getting it to work in the context we need, and haven't documented it yet, so WIP


src/Reflex/Main.hs
```diff
@@ -0,0 +1,69 @@
...
+import Data.Maybe (catMaybes)
+import Control.Monad.IO.Class
+import Control.Monad (forever)
+import Data.Traversable (for)
+import Data.Foldable (for_)
+import Data.Functor.Identity
+
+type EventChannel t = Chan [DSum (EventTriggerRef t) TriggerInvocation]
+
+runHost
```
ryantrinkle: Let's give this a slightly longer name, like `runHostWithIO`, to indicate that it grants access to IO.  I don't want people to get the impression that all hosts can/should support IO.

dalaing:
If you wanted to take anything from https://github.com/qfpl/reflex-basic-host/blob/master/src/Reflex/Host/Basic.hs for this, feel free.

It started life as almost exactly what you have here, then I needed the ability to sometimes use an Event to shutdown the host, which is where it started to diverge, and then I needed a newtype and the various instances to use this in a context where the pile-o'-constraints approach was giving me grief.

ryantrinkle:
Having an event to shut down reflex-dom apps would be great :)

dalaing:
I've been looking at adding that kind of event to as many of the reflex-dom backends as I can, for use with criterion and QuickCheck/Hedgehog...

ryantrinkle:
@dalaing It sounds like it would be great to base reflex-dom on that!

ali-abrar:
It would be nice if something like this also worked with reflex-vty.

ali-abrar:
https://github.com/reflex-frp/reflex-vty/blob/develop/src/Reflex/Vty/Host.hs#L89

Ericson2314:
OK Good CI works. Next step is to make reflex-vty use this.

endgame:
Is this still relevant given that `Reflex.Host.Headless` has been released?

ali-abrar:
I think the point of this was to provide something to make it easier to implement headless, vty, or other hosts.

PR #307, opened 2019-05-09
Add function rotation utils

alexfmpe:
In the same vein as #191, but these allow extending the `ffor2/ffor3` idiom to other functions, or other depths, which IME are both extremely common patterns:

```haskell
rotl2 attachWithMaybe (current xDyn) (updated yDyn) $ \x y -> ...big-block-of-code...
```
```haskell
rotl2 (liftA2 . liftA2) xDynOfEither yDynOfEither $ \x y -> ...big-block-of-code...
```

Right-rotations mostly included for symmetry

### DIFF
--- a/src/Reflex/Class.hs
+++ b/src/Reflex/Class.hs
@@ -149,6 +149,10 @@ module Reflex.Class
   , ffor
   , ffor2
   , ffor3
+  , rotl2
+  , rotl3
+  , rotr2
+  , rotr3
     -- * Deprecated functions
   , switchPromptly
   , switchPromptOnly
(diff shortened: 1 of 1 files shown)


3noch:
ffor2 is already "pushing it" IMO as to what should be in a core library like this. These functions definitely fall out of that if you ask me. When you include functions like this in a core library, you're saying "I want this to be common vocabulary for these apps" and I'm not sure I'd be comfortable making these "common vocabulary".

alexfmpe:
Yeah, I get what you mean. I also don't super like these combinators per-se.
Pondered a good bit on whether/where to open this PR, but ultimately didn't find a better place. There's no `reflex-contrib` package/repo equivalent of [reflex-dom-contrib](https://github.com/reflex-frp/reflex-dom-contrib).

dalaing:
I think `reflex-contrib` could be valuable in the same ways as `reflex-dom-contrib` has been - maybe it should become a thing?

ali-abrar:
Have you seen [composition-extra](https://hackage.haskell.org/package/composition-extra)? That seems to have the [rotation combinators](https://hackage.haskell.org/package/composition-extra-2.0.0/docs/Data-Function-Slip.html) you've got here, and some similar functions for [nested mapping] (https://hackage.haskell.org/package/composition-extra-2.0.0/docs/Data-Functor-Syntax.html). Functions like this more properly belong to a library like that, I suspect.

ali-abrar:
I agree, also, with @3noch's point. I'm not convinced that any of these need to be in the standard vocabulary. On the (I think, rare) occasions when something with this type signature would be useful, I would probably find the long-form much more readable than "rotl3". If I was working with code where this was an extremely common operation, I'd want to rethink _something_ or provide a local definition.

alexfmpe:
> Have you seen [composition-extra](https://hackage.haskell.org/package/composition-extra)? 
Huh, apparently I fail big time at hoogling. Thanks for bringing this to my attention

cgibbard:
Lambda exists for a reason.

3noch:
I prefer church encoded lambdas myself.

PR #310, opened 2019-05-30
fan: Rewrite haddock

endgame:
It appears that partial application for `fan` went away, but the haddock never changed (how can you partially apply a unary function?).

I have also removed the `TODO` comment, as it seemed no longer relevant. I can re-add it if required.


ryantrinkle:
Thanks!  Yes, the change to using `EventSelector` addresses these issues, I think.

PR #318, opened 2019-07-04
Generalize fan following DMap

treeowl:
Generalize `fan` to `fang` (yeah, that's a horrible name) allowing fanning `DMap`s with arbitrary targets (not just `Identity`). I haven't yet dug down enough to work out how to do the same for `merge`, but it would be useful.


ryantrinkle:
Awesome!  This all looks great; maybe we could rename it to `fanG` and merge?

treeowl:
That's fine by me. I don't really like that name either, but ... whatever.
Do you have any suggestions for tackling that for merge? It looks a lot
hairier, and I don't even know where to start.

treeowl:
Renamed. Should be ready to merge.

src/Reflex/Class.hs
```diff
@@ -298,6 +302,18 @@ class ( MonadHold t (PushM t)
   mergeIntIncremental :: Incremental t (PatchIntMap (Event t a)) -> Event t (IntMap a)
   fanInt :: Event t (IntMap a) -> EventSelectorInt t a
 
+-- | Efficiently fan-out an event to many destinations. You should save the
+-- result in a @let@-binding, and then repeatedly 'select' on the result to
+-- create child events
+fan :: forall t k. (Reflex t, GCompare k)
+    => Event t (DMap k Identity) -> EventSelector t k
+    --TODO: Can we help enforce the partial application discipline here?  The combinator is worthless without it
+fan e = EventSelector (fixup (selectg (fanG e) :: k a -> Event t (Identity a)) :: forall a. k a -> Event t a)
```
treeowl: Should we match on the `Coercion` and then produce a lambda, or something like that, to make sure there's only one `eventCoercion` call per `Reflex` instance?

treeowl: It probably doesn't actually matter...

Ericson2314: I was thinking about such things and opened https://github.com/reflex-frp/reflex/issues/319

cgibbard:
If we're going with fanG, probably best to have it be selectG for
consistency.

Also, it's worth thinking about whether it wouldn't be a more reasonable
option to just make this generalisation just be fan itself -- the number of
occurrences of the fully general 'fan' in the wild is pretty small, and
this is a straightforward generalisation.

ryantrinkle:
The deprecation plan shouldn't burn our more hardcore users, but it might
make sense to eventually come around to fan=fanG

Ericson2314:
We'll have the same thing again with NonEmptyDMap for fan and merge, so I say new name for now, and replace `fan` later.

ali-abrar:
Could you please update the changelog?

treeowl:
@ali-abrar, updated.

saulzar:
Generalize fan to fanG allowing fanning DMaps with functors other than Identity.

Issue #233, opened 2018-10-10
EventWriterT change 


Saulzar:
EventWriterT was recently changed so that it's `tellEvent`is now prompt with regards to runWithReplace, that is that if a `tellEvent e` co-occurs when it is switched out by the same event `e` then the tellEvent does not occur.

https://github.com/reflex-frp/reflex/commit/ca50bb4cf1e85e1c980b3ddad0e1b0c483542349#diff-ca34969f303705bc74cc027f0eedb19f

I would argue that in a lot of cases runWithReplace it would be desirable to have `tellEvent` be non promptly switched, does this require two kinds of EventWriterT one for prompt and one for non prompt switching?

Saulzar:
After more investigating I'm not even sure quite what the behaviour of the new changes is. I have a lot of uses of this pattern where I have some code which looks like this:

```
state = Workflow $ do
    ....
    tellEvent foo
    ...
    return (x, baseState <$ foo)

```

Most of them seem to continue working (contrary to the description of the new behaviour). One or two of them have stopped working and I'm not really sure why. Will see if I can find an isolated test case.

Saulzar:
FWIW. In one case I have this code:

```
performEvent (liftIO . print <$> e1)
tellEvent e1

```

This seems to only occur when `e1` is the result of a switch from some other event `e2` and `e2` is causing this widget to be replaced. So perhaps EventWriterT has the correct behavior here.
The performEvent prints, but tellEvent does not fire - at the least this is inconsistent?

JBetz:
Possibly related. since upgrading to `reflex-0.5`, a lot of my widgets that use `tellEvent` are broken. The event only gets sent up the first time it's triggered. 

For example:

```haskell
dyn_ $ ffor stateD $ \(State _) -> do
... 
tellEvent $ pure <$> select ZoomIn menuE
tellEvent $ pure <$> select ZoomOut menuE
...
```

I can only "tell" one of those events, and it can only be "told" once.

Is this related to the change you mentioned?

ryantrinkle:
@JBetz I believe @Saulzar's case would apply if `stateD` is changing at the same time that `menuE` is firing.  However, it doesn't explain why you'd be seeing only a single `tellEvent` taking effect.  Would you mind posting a bit more context? I'd like to see if I can reproduce the issue and get it fixed ASAP.

[...]

PR #320, opened 2019-07-08
Generalize merge

treeowl:
Generalize types of merge methods.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -3,6 +3,10 @@
 ## Unreleased
 
 * Fix `holdDyn` so that it is lazy in its event argument
+* Generalize `merge` to `mergeG`, `mergeIncremental` to
+  `mergeIncrementalG`, and `mergeIncrementalWithMove`
+  to `mergeIncrementalWithMoveG`. These produce `DMap`s
+  whose values needn't be `Identity`.
 
 ## 0.6.1.0
 
--- a/reflex.cabal
+++ b/reflex.cabal
@@ -48,6 +48,7 @@ library
     data-default >= 0.5 && < 0.8,
     dependent-map >= 0.2.4 && < 0.3,
     exception-transformers == 0.4.*,
+    profunctors,
     lens >= 4.7 && < 5,
     monad-control >= 1.0.1 && < 1.1,
     monoidal-containers == 0.4.*,
--- a/src/Reflex/Class.hs
+++ b/src/Reflex/Class.hs
@@ -17,6 +17,7 @@
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE TypeOperators #-}
 {-# LANGUAGE UndecidableInstances #-}
+{-# LANGUAGE Trustworthy #-}
 #ifdef USE_REFLEX_OPTIMIZER
 {-# OPTIONS_GHC -fplugin=Reflex.Optimizer #-}
 #endif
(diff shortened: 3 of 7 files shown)


Ericson2314:
It might still be possible to push the function so it is not exposed in the class? Then it could still have @saulzar's signature? I do think that is better for future compatibility.

treeowl:
What does "push the function" mean?

treeowl:
By the way: if the concern is the performance characteristics, I think we *might* be able to do something like

```haskell
mergeG :: GCompare k => (forall a. Coercible (q a) (Event t (v a)))
         => DMap k q -> Event t (DMap k v)
```

It's not very pretty, but I think it'll probably let you use both `Compose (Event t) Identity` and `Event t`. I haven't tried it though.

Ericson2314:
@treeowl sorry about my hastily-written comment. I edited it to be hopefully be clear. I am just concerned about not wanted/needing that parameter if we get something like https://github.com/ghc-proposals/ghc-proposals/pull/248.

treeowl:
Sorry, "push the application to the deeper" doesn't mean anything to be either. Can you show what you mean?

Ericson2314:
I started playing with the code, and then I realized I was being stupid; I was thinking "oh you use the same argument every time", but of course the general use is not `coerceEvent`.

Also playing I around I noticed damn this is way more annoying then even when I was making my examples in #233 days ago. I thought it was GHC 8.4 but 8.6 was no better. It seems quite separate from future proposals, the special `Constraint` solver stuff is downright buggy with local reasoning. I tried replacing `Event t` with an abstract `e` to get it to stop, but it was of no avail. I now see why `gcoerceWith` and pattern matching on `Coercion` has been useless so far. What a mess!

Given all that, great work! I like your
```haskell
mergeG :: GCompare k => (forall a. Coercible (q a) (Event t (v a)))
         => DMap k q -> Event t (DMap k v)
```
perhaps that could be an immediately-deprecated `mergeG0`, than then @saulzar's `mergeG` is written in terms of it as is `merge`? Maybe I'm too bullish on something being better in 8.8 or 8.10, but this status quo is just so awful I can hardly stand coding around it.

treeowl:
No, `mergeG0` doesn't work either. We can't satisfy `(forall a. Coercible (q a) (Event t (v a)))`. The best we can do in that regard is `(forall a. (Coercion (q a) (Event t (v a))))`, coercing with that everywhere the current PR calls for `nt`.

treeowl:
I'm really not a fan of `mergeG0` anyway. What makes passing functions un-reflexish?

saulzar:
I can confirm performance is identical (when patched against the same version of reflex) with this (unsurprisingly perhaps!).

Ericson2314:
For me it's not reflex specific, just question a question of 1) does the higher order function increase expressive power 2) allow some more performant hand-fusing. `traverse` and `>>=` don't pass 1) when there's `sequence` and `join`, for example, but do pass 2).

The naive reflex implementation's `mergeG` doesn't pass 2) since it itself maps the function making an intermediate `DMap`, but it looks like Spider's does pass? If so, I take back my criticism: the HOF is a good idea even were coercions not so underpowered.

treeowl:
@Ericson2314 the `Pure` implementation does indeed map over the `DMap`, but *it did so before too*. Now it just maps something else over it at the same time. I have very little sense of the performance implications for `Spider` when the function is non-trivial, because I'm still just starting to learn about what `Spider` gets up to.

Ericson2314:
+1 OK that is a good sign too wrt hand fusion.

PR #321, opened 2019-07-09
Use less unsafeCoerce

treeowl:
Get rid of a large fraction of calls to `unsafeCoerce`. Several
of the ones that remain are removed in my merge generalization
PR.

### DIFF
--- a/src/Reflex/Class.hs
+++ b/src/Reflex/Class.hs
@@ -32,6 +32,7 @@ module Reflex.Class
   ( module Reflex.Patch
     -- * Primitives
   , Reflex (..)
+  , Incremental (..)
   , mergeInt
   , coerceBehavior
   , coerceEvent
@@ -235,7 +236,7 @@ class ( MonadHold t (PushM t)
   -- Instead of always fully replacing the value, only parts of it can be patched.
   -- This is only needed for performance critical code via `mergeIncremental` to make small
   -- changes to large values.
-  data Incremental t :: * -> *
+  data Incremental' t :: * -> * -> *
   -- | A monad for doing complex push-based calculations efficiently
   type PushM t :: * -> *
   -- | A monad for doing complex pull-based calculations efficiently
@@ -296,8 +297,13 @@ class ( MonadHold t (PushM t)
   -- | Construct a 'Coercion' for a 'Dynamic' given an 'Coercion' for its
   -- occurrence type
   dynamicCoercion :: Coercion a b -> Coercion (Dynamic t a) (Dynamic t b)
+  -- | Construct a 'Coercion' for an 'Incremental'' given an 'Coercion' for its
+  -- occurrence type
+  incrementalCoercion :: Coercion t1 t2 -> Coercion p1 p2 -> Coercion (Incremental' t t1 p1) (Incremental' t t2 p2)
   mergeIntIncremental :: Incremental t (PatchIntMap (Event t a)) -> Event t (IntMap a)
   fanInt :: Event t (IntMap a) -> EventSelectorInt t a
+
+newtype Incremental t p = Incr { getIncr :: Incremental' t (PatchTarget p) p }
 
 --TODO: Specialize this so that we can take advantage of knowing that there's no changing going on
 -- | Constructs a single 'Event' out of a map of events. The output event may fire with multiple
--- a/src/Reflex/Profiled.hs
+++ b/src/Reflex/Profiled.hs
@@ -45,6 +45,7 @@ import Reflex.PerformEvent.Class
 
 import System.IO.Unsafe
 import Unsafe.Coerce
+import Data.IntMap (IntMap)
 
 data ProfiledTimeline t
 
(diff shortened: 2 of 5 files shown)


src/Reflex/Class.hs
```diff
@@ -296,9 +297,14 @@ class ( MonadHold t (PushM t)
   -- | Construct a 'Coercion' for a 'Dynamic' given an 'Coercion' for its
   -- occurrence type
   dynamicCoercion :: Coercion a b -> Coercion (Dynamic t a) (Dynamic t b)
+  -- | Construct a 'Coercion' for an 'Incremental'' given an 'Coercion' for its
+  -- occurrence type
+  incrementalCoercion :: Coercion t1 t2 -> Coercion p1 p2 -> Coercion (Incremental' t t1 p1) (Incremental' t t2 p2)
```
Ericson2314: I take it something goes wrong if you do this instead?
```haskell
incrementalCoercion :: Coercion (PatchTarget p1) (PatchTarget p2) -> Coercion p1 p2 -> Coercion (Incremental t p1) (Incremental t p2)
```

src/Reflex/Profiled.hs
```diff
@@ -144,16 +145,32 @@ instance Reflex t => Reflex (ProfiledTimeline t) where
...
-    Coercion -> unsafeCoerce (Coercion :: Coercion (Event (ProfiledTimeline t) a) (Event (ProfiledTimeline t) a)) --TODO: Figure out how to make this typecheck without the unsafeCoerce
-  dynamicCoercion (c :: Coercion a b) = case dynamicCoercion c :: Coercion (Dynamic t a) (Dynamic t b) of
-    Coercion -> unsafeCoerce (Coercion :: Coercion (Dynamic (ProfiledTimeline t) a) (Dynamic (ProfiledTimeline t) a)) --TODO: Figure out how to make this typecheck without the unsafeCoerce
-  mergeIntIncremental = Event_Profiled . mergeIntIncremental . (unsafeCoerce :: Incremental (ProfiledTimeline t) (PatchIntMap (Event (ProfiledTimeline t) a)) -> Incremental t (PatchIntMap (Event t a)))
+  currentIncremental (Incr (Incremental_Profiled i)) = coerce $ currentIncremental (Incr i)
+  updatedIncremental (Incr (Incremental_Profiled i)) = coerce $ profileEvent $ updatedIncremental (Incr i)
+  incrementalToDynamic (Incr (Incremental_Profiled i)) = coerce $ incrementalToDynamic (Incr i)
+  behaviorCoercion (c :: Coercion a b) =
+    Coercion `trans` (behaviorCoercion c :: Coercion (Behavior t a) (Behavior t b)) `trans` Coercion
```
Ericson2314: Can you do
```haskell
case fooCoercion c :: ... of Coercion -> Coercion
```
for all of these to make GHC work more?

src/Reflex/Class.hs
treeowl: That's a very good question. That probably will work, and it's much less horrible.

src/Reflex/Profiled.hs
treeowl: Can you give an example? Facing these somewhat involved coercion typed that I couldn't prove directly with `Coercion`, I reached for the toolbox. It should certainly be possible to do it all with `case`, but I don't know if it'll be any nicer. You're more than welcome to try!

src/Reflex/Profiled.hs
```diff
@@ -144,16 +145,32 @@ instance Reflex t => Reflex (ProfiledTimeline t) where
   unsafeBuildIncremental (ProfiledM a0) (Event_Profiled a') = coerce $ unsafeBuildIncremental a0 a'
   mergeIncremental = Event_Profiled . mergeIncremental . (unsafeCoerce :: Incremental (ProfiledTimeline t) (PatchDMap k (Event (ProfiledTimeline t))) -> Incremental t (PatchDMap k (Event t)))
   mergeIncrementalWithMove = Event_Profiled . mergeIncrementalWithMove . (unsafeCoerce :: Incremental (ProfiledTimeline t) (PatchDMapWithMove k (Event (ProfiledTimeline t))) -> Incremental t (PatchDMapWithMove k (Event t)))
```
Ericson2314: I think these can be derived with:
```haskell
case (incrementalCoercion, eventCoercion) of
  (Coercion, Coercion) -> Coercion
```
I'd certainly hope the newtype coercion axioms are defined for data families too!

src/Reflex/Profiled.hs
treeowl: Oh, I think I see what you mean now. `Coercion . fancy . Coercion` should be `case fancy of Coercion -> Coercion`. But it seems potentially less maintainable.

Ericson2314: Yeah that's what I mean. I was thinking *more* maintainable, since GHC can figure different things out there as needed. I guess I'm also bullish in GHC becoming smarter in the future. (I'm planning on writing a proposal for getting rid of roles.)

treeowl:
I'll have to work through this again tomorrow.

src/Reflex/Spider/Internal.hs
```diff
@@ -935,17 +941,41 @@ type BehaviorEnv x = (Maybe (Weak (Invalidator x), IORef [SomeBehaviorSubscribed
 
 --type role BehaviorM representational
 -- BehaviorM can sample behaviors
-newtype BehaviorM x a = BehaviorM { unBehaviorM :: ReaderT (BehaviorEnv x) IO a } deriving (Functor, Applicative, MonadIO, MonadFix)
```
Ericson2314: What is wrong with `ReaderT`? The role of it's first parameter is representational. `BehaviorEnv`'s one parameter is phantom. I would think that GHC could thread the phantom fine?

treeowl: The trouble is the third parameter, which is nominal. :-(

Ericson2314: Ah right. Thanks.

treeowl: We could generalize to something like `RIO`, and probably should do so internally. As you suggest, there's no need to pin down the environment type too eagerly. Only the monad needs to be fixed.

ryantrinkle: Why is the 3rd parameter nominal?  Is this a property of ReaderT?

Ericson2314: @ryantrinkle the third type parameter of `ReaderT` is nominal only because roles suck: the constraint one wants is `(Coerciable r0 r1, Coerciable (m0 a0) (m1 a1)) => Coerciable (ReaderT r0 m0 a0) (r1 m1 a1))`.

treeowl:
Closing in favor of #322, based on @Ericson2314's suggestion regarding `Dynamic`.

PR #322, opened 2019-07-09
Less unsafeCoerce

treeowl:
Get rid of most unnecessary uses of `unsafeCoerce`. Several of the ones that remain are removed by #320.

### DIFF
--- a/reflex.cabal
+++ b/reflex.cabal
@@ -56,6 +56,7 @@ library
     mtl >= 2.1 && < 2.3,
     prim-uniq >= 0.1.0.1 && < 0.2,
     primitive >= 0.5 && < 0.7,
+    profunctors,
     random == 1.1.*,
     ref-tf == 0.4.*,
     reflection == 2.1.*,
@@ -71,6 +72,7 @@ library
     witherable >= 0.2 && < 0.4
 
   exposed-modules:
+    Control.Monad.ReaderIO
     Data.AppendMap,
     Data.FastMutableIntMap,
     Data.FastWeakBag,
--- a/src/Control/Monad/ReaderIO.hs
+++ b/src/Control/Monad/ReaderIO.hs
@@ -1,6 +1,7 @@
 {-# language RoleAnnotations #-}
 {-# language MultiParamTypeClasses #-}
 {-# language FlexibleInstances #-}
+{-# language CPP #-}
 module Control.Monad.ReaderIO
   (
     ReaderIO (..)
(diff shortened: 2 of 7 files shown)


treeowl:
Big question: how should I name the `RIO` module? I don't want it getting mixed up with the one from `rio`, since it's pretty much internal. I don't care myself.

treeowl:
@Ericson2314 I was quite unable to get anything good out of the manual `case` on `Coercion` idea. For whatever reason, I couldn't get GHC to infer the right coercion flow. But I was able to dramatically reduce the type annotations I was using before.

Ericson2314:
@treeowl I think you forget to add the RIO module. OTOH I think there might be another way around the `RIO` stuff.

treeowl:
@Ericson2314 Whoops, fixed that. The only way I see around `RIO` is to do something like `DynamicS x p = Dynamic x (Behavior x (PatchTarget p)) p`. That might do the trick, but it doesn't exactly improve clarity.

Ericson2314:
@treeowl Yeah and also if other roles depend on `BehaviorM` their's no opportunity to provide your own coercion to prove a non-nominal role.

cgibbard:
Wait, what? I see exposed-modules: Control.Monad.RIO, and I see an import
of that module, but I don't see that module itself.

Ericson2314:
@cgibbard see I wrote that above :)  @treeowl I think you meant to push but didnt'?

treeowl:
Oh crud. Yeah, I forgot to push. Actually fixed now, I believe!

ryantrinkle:
To avoid confusion with https://github.com/commercialhaskell/rio/blob/10d0bd8c6b03035504c5c086dc90ad85cdb0b914/rio/src/RIO/Prelude/RIO.hs#L38, I think it would be better for us to choose an uglier name for our reader-over-IO.  In particular, our goals are pretty different: RIO is aimed at improving the elegance of user code, whereas this reader-over-IO is simply a hack around the role system's inability to express quite the right thing.  Given that one can't be substituted for the other (RIO doesn't fix the role issue, and this thing doesn't help with any of RIO's goals), I think it would be better to use a different, uglier name.  I'd propose `ReaderT_IO` or `ReaderIO` just so it's blatant what's going on.  Ideally, if GHC improves the role system in the future, this would go away.

treeowl:
@ryantrinkle I've renamed the module and type as you requested.

PR #324, opened 2019-07-10
Touch after, not before

treeowl:
Touching a heap object and then performing an operation is
weird; the garbage collector could run in between. Unless
these functions are used in a surprising context, it seems
most likely that they should change.


treeowl:
This was an XY problem. See #325.

PR #325, opened 2019-07-11
Better unsafe perform io

treeowl:
The previous code was struggling to create artificial dependencies
to prevent `unsafePerformIO` calls from being floated out of their
proper context. Push the context into each `unsafePerformIO` call
so the dependency is real. Reading a bit of the resulting Core,
I don't expect to see many (if any) performance regressions.
Performance will hopefully improve a bit if/when
[GHC issue 15127](https://gitlab.haskell.org/ghc/ghc/issues/15127)
is fixed.

### DIFF
--- a/src/Reflex/Spider/Internal.hs
+++ b/src/Reflex/Spider/Internal.hs
@@ -192,12 +192,6 @@ nextNodeIdRef = unsafePerformIO $ newIORef 1
 
 newNodeId :: IO Int
 newNodeId = atomicModifyIORef' nextNodeIdRef $ \n -> (succ n, n)
-
-{-# NOINLINE unsafeNodeId #-}
-unsafeNodeId :: a -> Int
-unsafeNodeId a = unsafePerformIO $ do
-  touch a
-  newNodeId
 #endif
 
 --------------------------------------------------------------------------------
@@ -301,9 +295,10 @@ cacheEvent e =
 #else
   Event $
 #endif
-    let mSubscribedRef :: IORef (FastWeak (CacheSubscribed x a))
-        !mSubscribedRef = unsafeNewIORef e emptyFastWeak
-    in \sub -> {-# SCC "cacheEvent" #-} do
+  unsafePerformIO $ do
+    mSubscribedRef :: IORef (FastWeak (CacheSubscribed x a))
+        <- newIORef emptyFastWeak
+    pure $ \sub -> {-# SCC "cacheEvent" #-} do
 #ifdef DEBUG_TRACE_EVENTS
           unless (BS8.null callSite) $ liftIO $ BS8.hPutStrLn stderr callSite
 #endif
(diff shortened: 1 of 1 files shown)


treeowl:
I believe it should be safe to use `unsafeDupablePerformIO` rather than `unsafePerformIO`. What's less clear to me is whether that's a good idea for performance; feel free to experiment!

src/Reflex/Spider/Internal.hs
```diff
@@ -1537,13 +1532,16 @@ getFanSubscribed k f sub = do
...
+      nid <- liftIO newNodeId
+#endif
       let subscribed = FanSubscribed
             { fanSubscribedCachedSubscribed = fanSubscribed f
             , fanSubscribedOccurrence = occRef
             , fanSubscribedParent = subscription
             , fanSubscribedSubscribers = subscribersRef
 #ifdef DEBUG_NODEIDS
-            , fanSubscribedNodeId = unsafeNodeId f
+            , fanSubscribedNodeId = nid
```
treeowl: Er... I'm not sure if this is right. the previous code only made the node ID depend on `f`; this makes it depend on more. What's semantically right?

src/Reflex/Spider/Internal.hs
```diff
@@ -1615,7 +1616,7 @@ getSwitchSubscribed s sub = do
             , switchSubscribedCurrentParent = subscriptionRef
             , switchSubscribedWeakSelf = weakSelf
 #ifdef DEBUG_NODEIDS
-            , switchSubscribedNodeId = unsafeNodeId s
+            , switchSubscribedNodeId = nid
```
treeowl: The same question applies here.

src/Reflex/Spider/Internal.hs
```diff
@@ -1670,7 +1674,7 @@ getCoincidenceSubscribed c sub = do
             , coincidenceSubscribedInnerParent = innerSubdRef
             , coincidenceSubscribedWeakSelf = weakSelf
 #ifdef DEBUG_NODEIDS
-            , coincidenceSubscribedNodeId = unsafeNodeId c
+            , coincidenceSubscribedNodeId = nid
 #endif
```
treeowl: And here.

src/Reflex/Spider/Internal.hs
treeowl: That is: do you *intend* the call to `unsafeNodeId` to float out? If so, how far?

src/Reflex/Spider/Internal.hs
```diff
@@ -1170,18 +1165,12 @@ buildDynamic readV0 v' = do
   return d
 
 unsafeBuildDynamic :: BehaviorM x (PatchTarget p) -> Event x p -> Dyn x p
-unsafeBuildDynamic readV0 v' = Dyn $ unsafeNewIORef x $ UnsafeDyn x
-  where x = (readV0, v')
+unsafeBuildDynamic readV0 v' =
+  Dyn $ unsafePerformIO $ newIORef $ UnsafeDyn (readV0, v')
```
treeowl: If this gets called *many times*, then it might be worth working around GHC 15127 by using `runRW#` manually to unbox the `IORef`.

```haskell
unsafeNewIORef :: a -> IORef a
unsafeNewIORef a = IORef $ STRef ( runRW# $ \s ->
  case newMutVar# a (noDuplicate# s) of
    (# _s', mv #) -> mv )
```

or

```haskell
unsafeDupableNewIORef :: a -> IORef a
unsafeDupableNewIORef a = IORef $ STRef ( runRW# $ \s ->
  case newMutVar# a s of
    (# _s', mv #) -> mv )
```

src/Reflex/Spider/Internal.hs
saulzar: It seems fine to me, I can't think why having the NodeId in it's own unsafePerformIO would matter - just for convenience and having one less 
#ifdef DEBUG_NODEIDS I am guessing?

treeowl: @saulzar it's just a question of what the node ID is supposed to identify.

ryantrinkle: Ideally I would like a 1-to-1 correspondence between node IDs and the cache IORefs that are being allocated.

src/Reflex/Spider/Internal.hs
saulzar: I don't think it does - because unsafeBuildDynamic is not a very frequently used function. Current benchmarks show no significant difference (for any of these changes)

PR #315, opened 2019-06-21
Add 'runWorkflow'

alexfmpe:

PR #326, opened 2019-07-13
Add missing details to documentation of 'networkView' and 'workflowView'

alexfmpe:
I've been using `dyn`/`networkView` for two years without realizing it'd fire the initial value at post-build time because I never read the actual source. Caused me to accidentally change semantics in https://github.com/reflex-frp/reflex/pull/315


src/Reflex/Network.hs
```diff
@@ -20,7 +20,7 @@ import Reflex.NotReady.Class
 import Reflex.PostBuild.Class
 
 -- | A 'Dynamic' "network": Takes a 'Dynamic' of network-creating actions and replaces the network whenever the 'Dynamic' updates.
--- The returned Event of network results fires when the 'Dynamic' updates.
+-- The returned Event of network results fires at post-build time and when the 'Dynamic' updates.
```
alexfmpe: Shouldn't this say network of Event instead?

PR #328, opened 2019-07-16
Patch maps better

treeowl:
Use the `containers` merge interface to improve efficiency.

Make strictness consistent.


treeowl:
I did a fancy useless-deletion detection thing for `Map`; I can do the same for `IntMap` if that's how we want to go. It would also be possible to *try* to detect useless updates, but it'll be pretty unreliable for many important cases. We could add a `PatchMap` type that insists on its values being `Eq`, but would that be useful?

src/Reflex/Patch/Map.hs
```diff
@@ -22,12 +25,34 @@ newtype PatchMap k v = PatchMap { unPatchMap :: Map k (Maybe v) }
...
+          (zipWithMaybeAMatched (\_k mv v -> Changed $! mv <|> Just v)) p old
+
+changedToMaybe :: Changed a -> Maybe a
+changedToMaybe (Unchanged _) = Nothing
+changedToMaybe (Changed a) = Just a
+
+data Changed a
+  = Unchanged a
+  | Changed a
+  deriving (Functor)
```
3noch: Are there any speed benefits to making this `newtype Changed a = Changed (ChangedTag, a)`? Or does that, as I imagine, compile down to the same thing?

treeowl: Should I change this to `data Changed a = Changed !Bool a` or similar? I'll have to look at Core to see if GHC will unbox it nicely in that case. Might be an improvement.

treeowl: See my comment from a minute ago.

cgibbard:
Haha, this kind of thing is exactly why I got you to make Data.Map.Merge in the first place, it's kind of funny that you'd also end up helping us replace our old awkward merges with it. :D

src/Reflex/Patch/Map.hs
lpsmith: Lazy pairs are kind of challenging to compile,  so yeah,  it's really between what @treeowl suggested and what's already written.  (But I don't think this issue should block anything on this MR.)

Ericson2314:
Moved to https://github.com/reflex-frp/patch/pull/3

PR #330, opened 2019-07-19
Add debug flag, make reflex compile with DEBUG set

saulzar:
- There were some antiquities with the cpp macro DEBUG, meaning it didn't compile. 
- An attempt to clean up the noise a little by abstracting some of the debug functions.


src/Reflex/Spider/Internal.hs
```diff
@@ -143,41 +130,9 @@ instance HasNodeId (Pull x a) where
...
-  EventNever -> "EventNever"
-  EventPush _ -> "EventPush"
-  EventMerge _ -> "EventMerge"
-  EventFan _ _ -> "EventFan"
-  EventSwitch _ -> "EventSwitch"
-  EventCoincidence _ -> "EventCoincidence"
-  EventHold _ -> "EventHold"
-  EventDyn _ -> "EventDyn"
-  EventHoldIdentity _ -> "EventHoldIdentity"
-  EventDynIdentity _ -> "EventDynIdentity"
```
saulzar: These are from a reflex of some time ago!

reflex.cabal
```diff
@@ -37,6 +37,12 @@ flag fast-weak
   default: False
   manual: True
 
+flag debug
+  description: Enable ebugging of spider internals
```
alexfmpe:
```suggestion
  description: Enable debugging of spider internals
```

saulzar:
@phadej  Might you have any idea where this "constraint from project config TODO requires installed instance" is coming from here? (or is this a red herring?)

I added a couple of flags and several versions fail in Travis, I can't seem to reproduce the problem locally.

Full error is 

```
[__5] rejecting: binary-0.8.5.1/installed-0.8... (conflict: reflex => containers>=0.6 && <0.7, binary => containers==0.5.11.0/installed-0.5...)
[__5] rejecting: binary-0.8.7.0, binary-0.8.6.0, binary-0.8.5.1, .... lots of versions .... binary-0.2, binary-0.10.0.0, binary-0.9.0.0 (constraint from project config TODO requires installed instance)
[__5] fail (backjumping, conflict set: binary, criterion, reflex)
```

.travis.yml
```diff
@@ -1,14 +1,13 @@
 # This Travis job script has been generated by a script via
 #
-#   haskell-ci '--config=cabal.haskell-ci' 'cabal.project'
+#   haskell-ci 'reflex.cabal'
```
phadej: The problem is that you don't use config file, that's why things go wrong in CI.

saulzar: Thanks!

test/DebugCycles.hs
```diff
@@ -78,7 +83,7 @@ addHeight e = leftmost [e4, e4] where
   e4 = leftmost [e3, e3]
 
 
--- Take an existing test and build it inside a push
+-- Take an existing test and build it inside a
```
alexfmpe: Inside a?

src/Data/FastMutableIntMap.hs
```diff
@@ -91,3 +92,6 @@ applyPatch (FastMutableIntMap r) p@(PatchIntMap m) = do
   v <- readIORef r
   writeIORef r $! applyAlways p v
   return $ IntMap.intersection v m
+
+toList :: FastMutableIntMap a -> IO [(Int, a)]
+toList (FastMutableIntMap r) = IntMap.toList <$> readIORef r
```
Ericson2314: Would `readToList`, `readList`, or `readAll` be better?

xplat: probably not actually; very little of the interface makes such distinctions.

src/Reflex/Spider/Internal.hs
```diff
@@ -111,46 +113,35 @@ import Reflex.PerformEvent.Base (PerformEventT)
 #ifdef DEBUG_TRACE_EVENTS
 import qualified Data.ByteString.Char8 as BS8
 import System.IO (stderr)
-#endif
-
-#ifdef DEBUG_TRACE_EVENTS
+import Data.List (isPrefixOf)
 
 withStackOneLine :: (BS8.ByteString -> a) -> a
```
xplat: probably would be better not to have the same ifdef spanning both imports and definitions, although that's not new to this PR

src/Reflex/Spider/Internal.hs
```diff
@@ -111,46 +113,35 @@ import Reflex.PerformEvent.Base (PerformEventT)
...
-
 #ifdef DEBUG
-
 #define DEBUG_NODEIDS
 
-debugPropagate = True
-
-debugInvalidateHeight = False
-
-debugInvalidate = True
```
xplat: i don't actually see the case for getting rid of these?

src/Reflex/Spider/Internal.hs
```diff
@@ -328,48 +299,69 @@ cacheEvent e =
...
+          liftIO $ cacheSubscription sub mSubscribedRef subscribedTicket
+
+cacheSubscription :: Subscriber x a -> IORef (FastWeak (CacheSubscribed x a))
+                  -> FastWeakTicket (CacheSubscribed x a) -> IO (EventSubscription x, Maybe a)
+cacheSubscription sub mSubscribedRef subscribedTicket = do
+  subscribed <- getFastWeakTicketValue subscribedTicket
+  ticket <- FastWeakBag.insert sub $ _cacheSubscribed_subscribers subscribed
+  occ <- readIORef $ _cacheSubscribed_occurrence subscribed
+
+  let parentSub = _cacheSubscribed_parent subscribed
```
xplat: This might change retention policy, we probably need to run it by Ryan to be sure it's okay

xplat: (but it's probably okay since we already retain subscribedTicket)

src/Reflex/Spider/Internal.hs
```diff
@@ -452,50 +447,40 @@ newSubscriberFan subscribed = return $ Subscriber
       _ <- DMap.traverseWithKey f $ DMap.intersectionWithKey (\_ -> Pair) a subs --TODO: Would be nice to have DMap.traverse_
       return ()
   , subscriberInvalidateHeight = \old -> do
-      when debugInvalidateHeight $ putStrLn $ "invalidateSubscriberHeight: SubscriberFan" <> showNodeId subscribed
```
xplat: we shouldn't actually need to lose these.

xplat: oh, i see, we don't ... but we have them on all the time now?  not sure that's great either ...

src/Reflex/Spider/Internal.hs
```diff
@@ -1378,20 +1405,51 @@ groupByHead = \case
       | otherwise -> (x, xs :| []) : l
 
 listsToForest :: Eq a => [[a]] -> Forest a
-listsToForest l = fmap (\(a, l') -> Node a $ listsToForest $ toList l') $ groupByHead $ catMaybes $ fmap nonEmpty l
+listsToForest lists = buildForest <$> groupByHead (mapMaybe nonEmpty lists)
+    where buildForest (a, lists') = Node a $ listsToForest $ toList lists'
+
+showStacks :: [[String]] -> String
+showStacks = drawForest . listsToForest . fmap (filterStack "Reflex.Spider.Internal")
```
xplat: is this really going to be as useful as it could be without user code in stacks?  in fact to figure out what reflex is doing i have sometimes had to capture extra stacks when nodes are created, never mind filtering things out ...

src/Reflex/Spider/Internal.hs
```diff
@@ -1869,59 +1962,71 @@ invalidateMergeHeight m = invalidateMergeHeight' (_merge_heightRef m) (_merge_su
...
+    -- currentHeight <- getCurrentHeight
+    -- when (height <= currentHeight) $ if height /= invalidHeight
+    --     then do
+    --       myStack <- liftIO $ whoCreatedIORef undefined --TODO
+    --       error $ "Height (" ++ show height ++ ") is not greater than current height (" ++ show currentHeight ++ ")\n" ++ unlines (reverse myStack)
+    --     else liftIO $
+    when (height == invalidHeight) $
+#ifdef DEBUG_CYCLES
+          do
+            nodesInvolvedInCycle <- walkInvalidHeightParents subscribed
```
xplat: might be nice to throw nodeIDs in with these stacks so you can relate them to propagation/invalidation traces from the same run ...

PR #333, opened 2019-07-25
TagMap replacement (WIP)

treeowl:
Separate concerns for `TagMap`, add functionality, and try to
manage the inherent unsafety a bit better. I don't really see
what `MyTag` has to offer by way of optimization, so I haven't
included that business. Maybe I'm missing something there...


414owen:
I think this is hitting https://github.com/ndmitchell/hlint/issues/818

Ericson2314:
Yes everything is going to hit that, so just ignore it for now. Trying to fix.

Ericson2314:
Let's do this after https://github.com/reflex-frp/reflex/pull/371

PR #335, opened 2019-08-04
haskell-ci for reflex

ali-abrar:
Thanks @phadej!


cabal.project
```diff
@@ -0,0 +1,13 @@
+packages: .
+
+-- why 0.3 is not in mokus0 tree?
```
ali-abrar: https://github.com/mokus0/dependent-map/pull/26 - should be merged and released shortly. I'll remove this override as soon as that's done.

ali-abrar: [It's up](https://hackage.haskell.org/package/dependent-map-0.3)

ali-abrar:
Closed in favor of #336 (which includes all of this branch)

PR #340, opened 2019-08-05
Fix merges

saulzar:
Augment missed changelog entries, add some missed version bounds


src/Reflex/Pure.hs
```diff
@@ -93,7 +93,8 @@ instance (Enum t, HasTrie t, Ord t) => Reflex (Pure t) where
        then Nothing
        else Just currentOccurrences
 
-  -- fanG :: GCompare k => Event (Pure t) (DMap k v) -> EventSelectorG (Pure t) k v
+--  The instance signature doeesn't compile, leave commented for documentation
```
alexfmpe: typo: doe**e**sn't

PR #345, opened 2019-08-10
Disable split-these by default (does not exist in reflex-platform)

saulzar:


ali-abrar:
Perhaps we can disable based on availability? Eg
https://github.com/bgamari/monoidal-containers/blob/master/default.nix#L14

saulzar:
That's a good idea, it seems to work (as far as when semialign does not exist). It seems like the previous implementation made it condition to a nix flag instead (is there such a thing, my understanding of nix is limited here?)

saulzar:
Actually I spoke too soon, this still gives me the familiar 

`error: anonymous function at /nix/store/hsbwq2mczs1ydvkf2g74bbclfvm73z7q-cabal2nix-reflex/default.nix:1:1 called without required argument 'semialign'`

ali-abrar:
Trying with this branch of reflex-platform might work: https://github.com/reflex-frp/reflex-platform/pull/516

saulzar:
Thanks! That gets further, also missing is these-lens. There doesn't seem to be compilers with the specific  versions i.e. ghc_8, ghc_8_4 etc. on the develop branch there however?

ali-abrar:
The platform only supports 8.6.5 at the moment. All prior versions have
been removed.

Ericson2314:
I rather just do https://github.com/reflex-frp/reflex/pull/364, which should be more maintainable.

Ericson2314:
No longer needed, cause we bumped the reflex platform used for CI.

PR #354, opened 2019-08-29
Fix broken recursive looping instance

saulzar:


ryantrinkle:
Does this actually fix it?  I'm confused about what the issue was

alexfmpe:
Nope, hence #355

PR #357, opened 2019-09-18
WIP add a couple of types for working with patches

danbornside:
Patchable generalizes PatchMap/SemiMap to any patchable type.  
PatchDMapWithReset allows to use patches in keys.  I didn't also solve the `WithMove` part since that adds significant typelevel overhead I didn't happen to need.


src/Reflex/Patch/Patchable.hs
```diff
@@ -0,0 +1,44 @@
+{-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE StandaloneDeriving #-}
+{-# LANGUAGE TypeFamilies #-}
+
+-- The derived instances are undecidable in the case of a pathlogical instance like
```
alexfmpe:
```suggestion
-- The derived instances are undecidable in the case of a pathological instance like
```

src/Reflex/Patch/DMapWithReset.hs
```diff
@@ -0,0 +1,105 @@
...
+  :: forall k v1 v2 v.
+  (GCompare k)
+  => (forall x. k x -> v1 x -> Maybe (v x))
+  -> (forall x. k x -> v2 x -> Maybe (v x))
+  -> (forall x. k x -> v1 x -> v2 x -> Maybe (v x))
+  -> DMap k v1 -> DMap k v2 -> DMap k v
+mergeWithKey f g fg = \xs ys -> DMap.mapMaybeWithKey onlyThat $ DMap.unionWithKey doIt (DMap.map This1 xs) (DMap.map That1 ys)
+  where
+    doIt _ (This1 xs) (That1 ys) = These1 xs ys
+    doIt _ _ _ = error "mergeWithKey misalligned keys"
```
3noch: `misaligned`

src/Reflex/Patch/DMapWithReset.hs
```diff
@@ -0,0 +1,120 @@
...
+
+import Data.Dependent.Map (DMap, GCompare (..))
+import qualified Data.Dependent.Map as DMap
+import Data.Semigroup (Semigroup (..))
+import Data.Constraint.Extras
+
+-- | A set of changes to a 'DMap'.  Any element may be inserted/updated or deleted.
+-- Insertions are represented as @'ComposeMaybe' (Just value)@,
+-- while deletions are represented as @'ComposeMaybe' Nothing@.
+newtype PatchDMapWithReset k p = PatchDMapWithReset { unPatchDMapWithReset :: DMap k (By p) }
```
saulzar: I don't see the connection with ComposeMaybe? Leftover from vanilla PatchDMap?

saulzar: The name "WithReset" - what is "Reset" in this case? 

I think there's a need for this type (also for normal Map), I've seen a few instances of this type in different places...

src/Reflex/Patch/DMapWithReset.hs
```diff
@@ -0,0 +1,120 @@
...
+          updates k ys = has @(Patches1Locally p) k $ \case
+            By_Insert x -> Just x
+            By_Delete -> Nothing
+            By_Patch x -> Just $ applyAlways x ys
+
+          inserts :: forall x. k x -> By p x -> Maybe (Patches1LocallyTarget p x)
+          inserts k = has @(Patches1Locally p) k $ \case
+            By_Insert x -> Just x
+            By_Delete -> Nothing
+            By_Patch _ -> Nothing
```
saulzar: What to do when the patch is invalid? Is it better to call error (which seems kind of ugly... )?

The use of Maybe as documented seems to be more about Patches which don't cause any updates, rather than for error handling - I guess this is the only sensible thing to do though?

src/Reflex/Patch/DMapWithReset.hs
3noch: I also didn't know what `Reset` meant. Even if it has some precedence somewhere out there, the fact that none of us understand it means we should probably change it.

src/Reflex/Patch/DMapWithReset.hs
danbornside: when is a patch invalid?

saulzar: For example when applying a patch "By_Patch" to a key which doesn't exist...

Ericson2314:
Superceded by https://github.com/reflex-frp/patch/pull/2

PR #358, opened 2019-10-08
Buncha instances

danbornside:


src/Reflex/Patch.hs
```diff
@@ -44,3 +52,39 @@ instance (Ord k, Group q) => Group (MonoidalMap k q) where
...
+instance Group (f (g a)) => Group ((f :.: g) a) where
+  negateG (Comp1 xs) = Comp1 (negateG xs)
+  Comp1 xs ~~ Comp1 ys = Comp1 (xs ~~ ys)
+
+instance Additive (f (g a)) => Additive ((f :.: g) a) where
+
+instance (Group (f a), Group (g a)) => Group ((f :*: g) a) where
+  negateG (a :*: b) = negateG a :*: negateG b
+  (a :*: b) ~~ (c :*: d) = (a ~~ c :*: b ~~ d)
+instance (Additive (f a), Additive (g a)) => Additive ((f :*: g) a) where
```
3noch: You can elide the `where` to make it clearer that you didn't intend to put anything here.

3noch:
We *could* just use the hackage package for this and get most of these: http://hackage.haskell.org/package/groups-0.4.1.0/docs/Data-Group.html

src/Reflex/Query/Class.hs
```diff
@@ -50,6 +55,25 @@ instance (Ord k, Query v) => Query (MonoidalMap k v) where
...
+  type QueryResult () = ()
+  crop _ _ = ()
+
+instance Query Void where
+  type QueryResult Void = ()
+  crop = absurd
+
+#if MIN_VERSION_base(4,12,0)
+-- | We can lift queries into monoidal containers.
+-- But beware of Applicatives who's monoid is different from (pure mempty, liftA2 mappend)
```
alexfmpe: whose?

Ericson2314:
Just need to fix GHC <= 8.2.

Ericson2314:
Ugh....hlint....

PR #359, opened 2019-10-20
Fix bug in `patchThatSortsMapWith`

johannesgerer:
Buggy implementation violates the invariant


JBetz:
I verified the fix by comparing the old behavior with the patch, and indeed, the old behavior breaks the invariant even though in both cases the patch can be applied with the same effect.

Test cases:
```haskell
main = do
  renderTest "OLD" patchThatSortsMapWith_old
  renderTest "NEW" patchThatSortsMapWith_new
  where
    renderTest title f = do
      let initial = Map.fromList $ zip ([0 ..] :: [Int]) "abc"
      let PatchMapWithMove patchMap = f (flip compare) initial
      let isValid = isJust $ patchMapWithMove patchMap
      let Just patched = apply (PatchMapWithMove patchMap) initial

      putStrLn title
      render "initial" initial
      render "patch" patchMap
      render "valid" isValid
      render "patched" patched
      putStrLn ""

    render k v =
      putStrLn $ k <> ": " <> show v
```

Outputs: 
```
OLD
initial: fromList [(0,'a'),(1,'b'),(2,'c')]
patch: fromList [(0,NodeInfo {_nodeInfo_from = From_Move 2, _nodeInfo_to = Just 0}),(2,NodeInfo {_nodeInfo_from = From_Move 0, _nodeInfo_to = Just 2})]
valid: False
patched: fromList [(0,'c'),(1,'b'),(2,'a')]

NEW
initial: fromList [(0,'a'),(1,'b'),(2,'c')]
patch: fromList [(0,NodeInfo {_nodeInfo_from = From_Move 2, _nodeInfo_to = Just 2}),(2,NodeInfo {_nodeInfo_from = From_Move 0, _nodeInfo_to = Just 0})]
valid: True
patched: fromList [(0,'c'),(1,'b'),(2,'a')]
```

Ericson2314:
@JBetz QA'd this, so we don't need separate QA. Looks good.

PR #360, opened 2019-10-21
Support semialign-1.1

phadej:


ryantrinkle:
Thanks!  It looks like the nix CI isn't any more broken than before, so I'll merge it despite the failure.

phadej:
Notice https://github.com/bgamari/monoidal-containers/pull/57 which prevents `reflex` to actually pick up `semialign-1.1` yet.

PR #366, opened 2019-11-25
Mention 'runWithReplace' in quickref

alexfmpe:
Took me over a year of using reflex to find out `runWithReplace` was a thing, and I have used workarounds due to not knowing about it in situations where I needed the initial value.


Ericson2314:
While `Adjustable` as a whole is somewhat gross, `runWithReplace` is super important and not really objectionable, so I agree with this.

CC @cgibbard as the chief `Adjustable` hater, what do you think?

PR #363, opened 2019-11-14
Modernize release.nix (includes #368)

Ericson2314:
This fixes CI, and tests more platforms (phones) for good measure.

PR #367, opened 2019-11-25
Change "traverse" to "traverse_" (includes #363)

matthewbauer:
In https://travis-ci.org/reflex-frp/reflex/builds/616304027 we notice
that hlint suggests:

```
   linting file src/Reflex/Spider/Internal.hs...
   reflex-0.6.2.4/src/Reflex/Spider/Internal.hs:497:13: Warning: Use traverse_
   Found:
     WeakBag.traverse (coincidenceSubscribedSubscribers subscribed)
       $ invalidateSubscriberHeight outerHeight
   Perhaps:
     traverse_ (coincidenceSubscribedSubscribers subscribed)
       $ invalidateSubscriberHeight outerHeight

   1 hint
```

This suggestion isn’t relevant because we don’t have a “traverse_”
equivalent for WeakBag.traverse. Looking at the code in
Hlint (src/Hint/Monad.hs), it becomes clear that it’s a dumb search
for any usage of “traverse”, regardless of being in the Traversable
typeclass.

What’s less clear is why this used to not be a problem before
recently. This error appears to coincide with an update of hlint from
v2.2.2 to v2.2.4.

To fix this, tell Hlint to ignore “Use traverse_” hints in Reflex.
This can be removed once Hlint is taught to distinguise between
traverse from the Traversable typeclass and from somewhere else.


ryantrinkle:
Maybe we should rename this to `traverse_`: https://github.com/reflex-frp/reflex/blob/develop/src/Data/FastWeakBag.hs#L124

ryantrinkle:
Do we have any other `traverse_` masquerading as `traverse`?

matthewbauer:
> Do we have any other `traverse_` masquerading as `traverse`?

It looks like the only one.

> Maybe we should rename this to traverse_: https://github.com/reflex-frp/reflex/blob/develop/src/Data/FastWeakBag.hs#L124

Yeah that sounds like a nice idea! I would assume we would want some kind of deprecation for traverse?

matthewbauer:
Updated with renaming traverse to traverse_.

Ericson2314:
(Merged in the other PR so we can fix all CI failures in one go so I am allowed to merge.)

PR #377, opened 2019-12-27
Fixes for ghc 8.8.1

hamishmack:
Moves `fail` from `Monad` instances when using ghc 8.8.1
and instead includes `MonadFail` where `fail` is needed.

Fixes #377

### DIFF
--- a/src/Reflex/Class.hs
+++ b/src/Reflex/Class.hs
@@ -200,7 +200,7 @@ import Data.Dependent.Map (DMap, DSum (..))
 import qualified Data.Dependent.Map as DMap
 import Data.Functor.Compose
 import Data.Functor.Product
-import Data.GADT.Compare (GEq (..), GCompare (..), (:~:) (..))
+import Data.GADT.Compare (GEq (..), GCompare (..))
 import Data.FastMutableIntMap (PatchIntMap)
 import Data.Foldable
 import Data.Functor.Bind
@@ -215,6 +215,7 @@ import Data.Some (Some(Some))
 import Data.String
 import Data.These
 import Data.Type.Coercion
+import Data.Type.Equality ((:~:) (..))
 import Data.Witherable (Filterable(..))
 import qualified Data.Witherable as W
 import Reflex.FunctorMaybe (FunctorMaybe)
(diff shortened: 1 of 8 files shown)


src/Reflex/Spider/Internal.hs
```diff
@@ -2637,6 +2641,10 @@ instance Monad (SpiderHost x) where
   SpiderHost x >> SpiderHost y = SpiderHost $ x >> y
   {-# INLINABLE return #-}
   return x = SpiderHost $ return x
+#if MIN_VERSION_base(4,13,0)
+
+instance MonadFail (SpiderHost x) where
```
Ericson2314: Can we define this before 8.8? Is `fail = MonadFail.fail` a default method?

src/Reflex/Class.hs
```diff
@@ -663,7 +663,6 @@ instance Reflex t => Monad (Behavior t) where
   a >>= f = pull $ sample a >>= sample . f
   -- Note: it is tempting to write (_ >> b = b); however, this would result in (fail x >> return y) succeeding (returning y), which violates the law that (a >> b = a >>= \_ -> b), since the implementation of (>>=) above actually will fail.  Since we can't examine 'Behavior's other than by using sample, I don't think it's possible to write (>>) to be more efficient than the (>>=) above.
   return = constant
-  fail = error "Monad (Behavior t) does not support fail"
```
Ericson2314: This needs CPP?

hamishmack:
Do we need `MonadFail` instances for things like `BehaviorM` when the underlying `ReaderIO` monad has no meaningful `fail`.  I think it is nicer to leave it out if possible and turn the runtime error (of calling `fail` for `BehaviorM`) into a compile time error?

hamishmack:
Removing the `MonadFail` instances that do not compile got things building again (I don't think we want them anyway). I have also switched to what I think is a newer dependent-sum.  Here is the [cabal.project](https://github.com/leksah/leksah/blob/ad7f71556b38794aa7c484798b4e70f0efd8046e/cabal.project#L32-L76) file I have been using for testing.

PR #379, opened 2020-01-08
Added semigroup constraint to (By p a) type

414owen:
To fix builds with ghc <= 8.2.2


3noch:
Does this produce a redundant constraint warning for newer GHCs?

Ericson2314:
Send that to https://github.com/reflex-frp/patch/pull/2 instead; also you should have push access to update the branch directly.

PR #382, opened 2020-01-09
Add changelog entry for 8.8 support

Ericson2314:
Not an interface change, but still notable.


Ericson2314:
hlint is wrong. Reporting upstream.

Ericson2314:
https://github.com/ndmitchell/hlint/issues/818

PR #383, opened 2020-01-09
Use patch library instead of Reflex.Patch.*

xplat:


Ericson2314:
I think we might have more things to report for https://github.com/ndmitchell/hlint/issues/818

Ericson2314:
Old failing CI for reference https://travis-ci.org/reflex-frp/reflex/builds/635019574?utm_source=github_status&utm_medium=notification

Ericson2314:
Yuck, I just have no idea how to fix this hlint thing, so merging anyways.

3noch:
Can't you just disable the lint in that file with an "hlint pragma"? See Ignoring Hints in https://hackage.haskell.org/package/hlint-1.7/src/hlint.htm

Ericson2314:
@3noch What's really annoying is that it works locally due to the freeze file banning the bad version.

PR #387, opened 2020-01-21
WIP: New workflows

alexfmpe:
For examples, see https://github.com/alexfmpe/obelisk/tree/ae%40test-new-workflows which currently has this code inlined for faster prototyping. Now that `ob run` supports loading dependencies into its repl, I'll switch to that instead.

The advantage of both `Wizard` and `Stack` over `Workflow` is that they can be composed much more easily. A `Workflow` declares all its future replacements up-front, while `Wizard` and `Stack` pieces can be created from a `m (Event t a)` and a life-cycle is then specified by chaining several widgets via bind or do-notation. AFAICT this is similar to [Concur](https://github.com/ajnsit/concur-documentation/blob/master/README.md#a-quick-tour-of-concur)'s `Widget`

Only one layer of a `Wizard` exists at any given time. Each time a wizard layer triggers its `Event`, it is completely replaced by the following one.
`Wizard` is adapted from https://github.com/reflex-frp/reflex/compare/workflow-monad. 

All layers of a `Stack` are present at the same time. Each time one layer triggers its `Event`, all following ones are replaced. A `Stack` can be made incremental with recourse to `MaybeT`.
`Stack` (or rather, something that looked similar but was completely bugged) was discovered by accident while trying to get away with type tetris and failing miserably.

The same temporal flow can be re-used for `Stack` and `Wizard`:
```haskell
let 
  someFlow mkWorkflow x0 = do
    x1 <- mkWorkflow x0
    x2 <- mkWorkflow x1
    x3 <- mkWorkflow x2
    pure x3

  wizardFlow = someFlow $
    step . someButton

  stackFlow = runMaybeT $ someFlow $
    MaybeT . frame . fmap (\ev -> (Nothing, fmap Just ev)) . someButton
```
To-do in no particular order:
- [ ] Add documentation (where do examples go?)
- [ ] Review typeclasses/constraints. I made a beeline for `Bind`/`Monad` and didn't pay much attention to the rest.
- [ ] Test for leaks
- [ ] Test for simultaneous occurrences
- [ ] Investigate relationship between `Workflow` and `Wizard`
- [ ] Try to make incremental stacks more ergonomic
- [ ] Try to make wizard/stack agnostic flows more ergonomic
- [ ] Try to make nesting of wizards/stacks/workflows more ergonomic

### DIFF
--- a/src/Reflex/Workflow.hs
+++ b/src/Reflex/Workflow.hs
@@ -1,5 +1,9 @@
+{-# LANGUAGE DeriveFunctor #-}
+{-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE RecursiveDo #-}
 {-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TemplateHaskell #-}
+{-# LANGUAGE TupleSections #-}
 -- |
 -- Module:
 --   Reflex.Workflow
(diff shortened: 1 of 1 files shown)


Ericson2314:
@alexfmpe I think we should deprecate `Workflow` in this library and do this stuff in a new downstream library. (`reflex-machines`, anyone?)

PR #389, opened 2020-02-04
Fix Reflex.Profiled infinite recursion for merge incremental

matthewbauer:
The Reflex Profiled transformer had an incorrect definition for
Reflex.Class.mergeIncrementalG and
Reflex.Class.mergeIncrementalWithMoveG. Instead of calling the
transformed inner Reflex timeline, it called itself, leading to an
infinite recursion. This is fixed by coercing the type before running
mergeIncremental*.

This was introduced in 8d8ceec, and
effects Reflex v0.6.2.1, v0.6.2.2, v0.6.2.3, v0.6.2.4, v0.6.3, and
v0.6.4.


Ericson2314:
The Travis failure is spurious, yet again.

PR #390, opened 2020-02-05
Add Reflex.Profiled test for CrossImpl

matthewbauer:
This uses Reflex.Profiled in addition to Pure & Spider for testing.
Reflex.Profiled wraps Spider here, so that it most results should be
the same. However, this should help to catch early some bad bugs that
occasionally occur in Reflex.Profiled.

### DIFF
--- a/test/Reflex/Test/CrossImpl.hs
+++ b/test/Reflex/Test/CrossImpl.hs
@@ -36,8 +36,6 @@ import Data.Traversable
 import System.Exit
 import System.Mem
 
-import System.IO.Unsafe
-
 mapToPureBehavior :: Map Int a -> Behavior PureReflexDomain a
 mapToPureBehavior m = P.Behavior $ \t -> case Map.lookupLE t m of
   Nothing -> error $ "mapToPureBehavior: no value for time " <> show t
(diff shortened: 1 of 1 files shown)


test/Reflex/Test/CrossImpl.hs
```diff
@@ -102,16 +106,19 @@ testAgreement :: (Eq c, Eq d, Show c, Show d) => (forall m t. TestCaseConstraint
 testAgreement builder inputs = do
   let identityResult = testPure builder inputs
   tracePerf "---------" $ return ()
-  let spiderResult = testSpider builder inputs
+  let spiderResult = unsafePerformIO $ S.runSpiderHost $ testTimeline builder inputs
   tracePerf "---------" $ return ()
-  let resultsAgree = identityResult == spiderResult
+  let profiledResult = unsafePerformIO $ S.runSpiderHost $ Prof.runProfiledM $ testTimeline builder inputs
```
3noch: For someone unfamiliar with this code like myself, I would need a nice comment explaining why `unsafePerformIO` is ok here.

3noch: Especially considering you have access to IO here?

matthewbauer: Yeah it shouldn't be necessary here.

PR #394, opened 2020-02-20
Quickref update align function documentation 

sambolias:
Only `alignEventWithMaybe` is exposed by [Reflex.Class module](https://github.com/reflex-frp/reflex/blob/develop/src/Reflex/Class.hs#L49). If `align` and `alignWith` were meant to be exposed that would be better fix, but I was looking to use them based on Quickref, and (unless I am missing something) this document is not reflecting what Reflex actually exposes for align functions. I was using reflex v0.63 brought in from reflex-platform v0.4.0.0, but it looks like adjacent versions are the same regarding this.

### DIFF
--- a/Quickref.md
+++ b/Quickref.md
@@ -50,14 +50,16 @@ Since MonadHold depends on MonadSample, any [S] function also runs in [H] contex
 [ ]   <@                         ::                        Behavior a -> Event b -> Event a
 
 -- Combine multiple Events
-[ ]   <>                  ::      Semigroup a => Event a -> Event a -> Event a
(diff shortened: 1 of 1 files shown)


3noch:
Oh these come from `Data.Align` in `semialign` package for which Event has an instance. But that should be clearer in the quickref.

sambolias:
I had looked in `Data.Align` in `align` package...I guess I gave up searching for that instance too quickly. Yes a note to that effect would have helped me a lot. Thank you.

3noch:
Thanks!

PR #399, opened 2020-03-16
Nits

endgame:
----

A couple of minor haddock and documentation fixes

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,5 +21,11 @@ A summary of Reflex functions is available in the [quick reference](Quickref.md)
 
 ### Hacking
 
-Use the `./scripts/hack-on reflex` script in [Reflex Platform](https://github.com/reflex-frp/reflex-platform) to checkout the source code of `reflex` locally in `reflex-platform/reflex` directory.
-Then do modifications to the source in place, and use the `./try-reflex` or `./scripts/work-on` scripts to create the shell to test your changes.
+From the root of a [Reflex
+Platform](https://github.com/reflex-frp/reflex-platform) checkout, run
+`./scripts/hack-on haskell-overlays/reflex-packages/dep/reflex`. This
+will check out the reflex source code into the
+`haskell-overlays/reflex-packages/dep/reflex` directory. You can then
+point that checkout at your fork, make changes, etc. Use the
+`./try-reflex` or `./scripts/work-on` scripts to start a shell in
+which you can test your changes.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -21,5 +21,10 @@ A summary of Reflex functions is available in the [quick reference](Quickref.md)
 
 ### Hacking
 
-Use the `./scripts/hack-on reflex` script in [Reflex Platform](https://github.com/reflex-frp/reflex-platform) to checkout the source code of `reflex` locally in `reflex-platform/reflex` directory.
-Then do modifications to the source in place, and use the `./try-reflex` or `./scripts/work-on` scripts to create the shell to test your changes.
+Use the `./scripts/hack-on
+haskell-overlays/reflex-packages/dep/reflex` script in [Reflex
```
cgibbard: Hmm, this instruction, though it may be accurate, has become a little awkwardly worded. Perhaps we should say something like "From the reflex-platform root directory, use the `./scripts/hack-on haskell-overlays/reflex-packages/dep/reflex` command to checkout the source code of reflex in `haskell-overlays/reflex-packages/dep/reflex`.

The directory in which hack-on will check out reflex will always be whatever directory you pass to it -- the one containing the thunk initially, so we should make sure that matches regardless.

This makes me wonder if reflex really should be buried so deep in reflex-platform. It makes sense to me for some of the thunks for core dependencies of reflex-platform perhaps to be slightly higher up in the directory structure -- maybe under a dep directory or something.

endgame: I've pushed a rewording. Please take a look.

README.md
```diff
@@ -21,5 +21,11 @@ A summary of Reflex functions is available in the [quick reference](Quickref.md)
 
 ### Hacking
 
-Use the `./scripts/hack-on reflex` script in [Reflex Platform](https://github.com/reflex-frp/reflex-platform) to checkout the source code of `reflex` locally in `reflex-platform/reflex` directory.
-Then do modifications to the source in place, and use the `./try-reflex` or `./scripts/work-on` scripts to create the shell to test your changes.
+From the root of a [Reflex
```
alexfmpe: Maybe just recommend the one-liner from #397 if we're not going with the wrapper script

endgame: I wasn't aware of #397 . Happy to withdraw this if we go for that instead.

endgame:
What's the decision re: #397 ? I am happy to split this PR and ship the haddock fix separately.

3noch:
CI actually passed for this but travis never updated the PR.

README.md
Ericson2314: In the future, I like 1 sentence per line because 1) no one is going to read this in a terminal 2) it makes the diffs saner.

PR #409, opened 2020-03-30
Add NOINLINE pragma to mkFastWeakTicket

tomsmalley:
(Also move what I think is the corresponding comment)

Without this NOINLINE, the FastWeakTicket can be fused away (in
cacheSubscription in particular), causing early unsubscriptions.

This manifested itself in reflex-dom-core as items in a nested list
never being added to the DOM, because runWithReplace in the immediate
builder was being unsubscribed from the lifted runWithReplace result
events.


Ericson2314:
Wow!

3noch:
This also explains why debugging code would make it go away!

Issue #414, opened 2020-04-30
Can `now :: PushM t (Event t ())` be implemented?


parenthetical:
I might be overlooking something, but does Reflex have a way to get an event occurrence at the current moment, e.g. through a value `now :: PushM t (Event t ())`? I'm asking mostly out of curiosity, although it might make what I'm working on  much simpler.

It is possible to get at it within `push` and then write a Reflex wrapper to have it available in every inner PushM, but that would leave the outer expression hanging:
```
pushWithNow :: (Reflex t)
            => (a -> Event t () -> PushM t (Maybe b))
            -> Event t a
            -> Event t b
pushWithNow f e = push (\a -> do
                           hE <- headE e
                           f a (() <$ hE))
                  e
```

ryantrinkle:
That seems like a very reasonable idea, and it should even admit a more efficient implementation than headE - in particular, you should never need to collect subscribers, because the event is known to be firing right away.

Would you like to take a stab at implementing it?  I think it will not be too bad, but I'm happy to support as necessary.

parenthetical:
I gave it a go in pull request #416 

Maybe there should also be a function `occurrence :: Event a -> PushM (Maybe a)`?
That would subsume `coincidence` with `push occurrence` (unless that messes with GC?), and it might avoid some awkward uses of alignEventWithMaybe.
It would also have the nice property of `occurrence =<< now   =   pure (Just ())`

PR #416, opened 2020-05-01
Extend the Reflex class with `now :: PushM t (Event t ())`. (Issue #414)

parenthetical:
I hope the implementation in Spider is correct, it produced the expected result when testing with `switchHoldPromptly never . pushAlways (\a -> fmap (a <$) now)`.

Extending the Reflex class might break modules depending on Reflex if they or another dependency use `now` as an identifier. Let me know if that's not okay.

The change is motivated by distributed Reflex (coming soon™), in which
you can write things like:
```
roundTrip :: (...) => Location t -> m (Event t ())
roundTrip to = headE =<< perceptionsHere =<< perceptionsAt to =<< now
```
Maybe there are other use cases as well?

### DIFF
--- a/src/Reflex/Class.hs
+++ b/src/Reflex/Class.hs
@@ -262,8 +262,6 @@ class ( MonadHold t (PushM t)
   never :: Event t a
   -- | Create a 'Behavior' that always has the given value
   constant :: a -> Behavior t a --TODO: Refactor to use 'pure' from Applicative instead; however, we need to make sure that encouraging Applicative-style use of 'Behavior's doesn't have a negative performance impact
-  -- | An event which occurs at the current moment.
-  now :: PushM t (Event t ())
   -- | Create an 'Event' from another 'Event'; the provided function can sample
   -- 'Behavior's and hold 'Event's, and use the results to produce a occurring
   -- (Just) or non-occurring (Nothing) result
@@ -422,6 +420,8 @@ class MonadSample t m => MonadHold t m where
   -- | Create a new 'Event' that only occurs only once, on the first occurrence of
   -- the supplied 'Event'.
   headE :: Event t a -> m (Event t a)
+  -- | An event which occurs at the current moment.
+  now :: m (Event t ())
 
 -- | Accumulate an 'Incremental' with the supplied initial value and the firings of the provided 'Event',
 -- using the combining function to produce a patch.
@@ -567,6 +567,7 @@ instance MonadHold t m => MonadHold t (ReaderT r m) where
   holdIncremental a0 = lift . holdIncremental a0
   buildDynamic a0 = lift . buildDynamic a0
   headE = lift . headE
+  now = lift now
 
 instance (MonadSample t m, Monoid r) => MonadSample t (WriterT r m) where
   sample = lift . sample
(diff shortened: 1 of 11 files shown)


ryantrinkle:
Overall this looks good.  Would you mind adding a new EventSubscribed, instead of reusing `eventSubscribedNever`?  They probably ought to be the same, except for, e.g. https://github.com/reflex-frp/reflex/blob/446094b733aa6d1e51418d0994889d2cc04fca43/src/Reflex/Spider/Internal.hs#L599

Also, we'll need some tests for this.

parenthetical:
I'll change that and try to add some tests as well, thanks!

parenthetical:
Does this look better?

ryantrinkle:
Can we add a test showing that it is *not* firing at some times other than the time it's created?  I think it might need an IORef or something - this looks like it is probably *always* firing (but only the first time each subscriber looks at it).

parenthetical:
Thanks for the hints; I implemented it with an IORef and added a test case for the always firing issue.
Was `scheduleClear` was the right procedure to use?

ryantrinkle:
Awesome!  That looks right to me :)

Only two things left:
1. You should add something to [the changelog](https://github.com/reflex-frp/reflex/blob/develop/ChangeLog.md) announcing this!
2. We'll take a look at what's going wrong with CI - I think that might be on our end.

parenthetical:
Great, thanks for the shepherding!
However, I just tried writing a program with it and now I'm not sure whether adding `now` to `Reflex` was the right thing to do, since I immediately needed a `HasNow` type class as well (which then has to use a different identifier ideally).

ryantrinkle:
Hmm, well, putting it in a typeclass seems like a perfectly reasonable thing to do.  Note that not all implementations of MonadSample or MonadHold will have implementations for `now`.  In particular: I think it'd be reasonable for people to assume that `hold a . (b <$) =<< now` will always update the Behavior to `b`, but that's not possible outside of a frame.  And, of course, `PullM` can't use `now` because cached results would be wrong.

src/Reflex/Class.hs
```diff
@@ -422,6 +420,8 @@ class MonadSample t m => MonadHold t m where
   -- | Create a new 'Event' that only occurs only once, on the first occurrence of
   -- the supplied 'Event'.
   headE :: Event t a -> m (Event t a)
+  -- | An event which occurs at the current moment.
+  now :: m (Event t ())
```
3noch, line 424: Perhaps it would make sense to add a default signature based on MonadTrans so that all these new methods can be added automatically.

parenthetical: Great idea, I've added it.

src/Reflex/Class.hs
```diff
@@ -420,6 +420,10 @@ class MonadSample t m => MonadHold t m where
...
```
3noch, line 424: Can any laws be documented?

parenthetical: I added this if that's what you're thinking of?
```
  -- | An event which only occurs at the current moment in time, such that:
  --
  -- > coincidence (pushAlways (\a -> (a <$) <$> now) e) = e
  --
  now :: m (Event t ())
```
Also, `now` turns events in the context of a `MonadHold` into an applicative functor and monad using the existing `Apply` and `Bind` instances:
```
instance (..., MonadHold t m) => Applicative (m . Event t) where
  pure a = (a <$) <$> now
  (<$>) = liftA2 (<.>)
```

ryantrinkle:
Awesome, thanks!

PR #429, opened 2020-05-24
Implement publish-subscribe pattern

JBetz:
Implements the [publish-subscribe pattern](https://en.wikipedia.org/wiki/Publish%E2%80%93subscribe_pattern) for communication within a reflex network. 

The motivation for this came from frustration with communication patterns being tightly coupled to widget hierarchies, particularly when implementing something similar to the Elm architecture by way of `EventWriter`. `PubSub` lets you separate these concerns, even inverting the widget hierarchy if so desired, as well as enabling a much more flexible event-handling architecture.

So far the only testing has been in application usage, but it builds upon existing abstractions without introducing anything too alien to reflex. I'm open to writing some documentation with examples if people are interested, but for now, just gauging interest and looking for feedback on merge-ability.


src/Reflex/PubSub.hs
```diff
@@ -0,0 +1,44 @@
...
+
+import Control.Monad.Reader
+import Data.Constraint.Extras
+import Data.Dependent.Map
+import Data.Dependent.Sum
+import Data.Functor.Identity
+
+import Reflex
+
+type PubSubT t tag m = EventWriterT t [DSum tag Identity] (ReaderT (Event t (DMap tag Identity)) m)
```
endgame, line 22: Suggest making this a `newtype`, so that you can define lifting instances for all the other useful classes, and people can use their own `EventWriter` or whatever else while pubsubbing.

That probably means a `Reflex.PubSub.Base` and `Reflex.PubSub.Class` pair of modules, for consistency with the rest of the library.

endgame: Should this perhaps be a `DList (DSum tag Identity)`? I'm thinking list is probably fine, as to have it go wrong, you'd need to have a lot of `payloadE` events firing in the same frame.

src/Reflex/PubSub.hs
```diff
@@ -0,0 +1,44 @@
...
+
+-- | 'PubSub' enables you to broadcast events from anywhere in your reflex network to
+-- any location that is subscribed to the topic they are published to. Topics are best
+-- representated as constructors of some custom GADT which carries a phantom type to
+-- constrain the messages that can be published to it.
```
endgame, line 27: Maybe add a note that this is consciously allowing spooky action-at-a-distance? (Not as badly as some other architectures, because you've tracked it with the type, but it's there.) This can be a fine and valid design choice, but ideally you want to `runPubSubT` over the minimal section of code that you want to talk to each other? (Like how it can be helpful to minimise the size of `rec` blocks.)

src/Reflex/PubSub.hs
```diff
@@ -0,0 +1,44 @@
...
+class (Reflex t, Monad m, Semigroup a) => PubSub t tag m a where
+  publish :: tag a -> Event t a -> m ()
+  subscribe :: tag a -> m (Event t a)
```
endgame, line 30: By analogy with `Control.Monad.Reader.asks`, maybe a `subscribes :: tag a -> (a -> b) -> m (Event t b)`? Could be useful with lens, where you can use `view someLens` as the function arg.

src/Reflex/PubSub.hs
```diff
@@ -0,0 +1,44 @@
...
+import Reflex
+
+type PubSubT t tag m = EventWriterT t [DSum tag Identity] (ReaderT (Event t (DMap tag Identity)) m)
+
+-- | 'PubSub' enables you to broadcast events from anywhere in your reflex network to
+-- any location that is subscribed to the topic they are published to. Topics are best
+-- representated as constructors of some custom GADT which carries a phantom type to
+-- constrain the messages that can be published to it.
+class (Reflex t, Monad m, Semigroup a) => PubSub t tag m a where
+  publish :: tag a -> Event t a -> m ()
```
endgame, line 29: Maybe add `publishes :: tag a -> (b -> a) -> Event t b -> m ()`? I'm less confident about this than I am `subscribes`, below, but it could be useful with lens: if you have `a ~ Endo c`, then you can use this to roll up a bunch of updates (generated from `Endo $ set someLens someVal`, for instance) and ship them around to different parts of your program? (Though maybe that's an anti-pattern, and the correct thing to do in that instance would be to stick a `Dynamic t` in a `ReaderT` instead of reconstructing it at each leaf?)

src/Reflex/PubSub.hs
```diff
@@ -0,0 +1,44 @@
...
+
+import Reflex
+
+type PubSubT t tag m = EventWriterT t [DSum tag Identity] (ReaderT (Event t (DMap tag Identity)) m)
+
+-- | 'PubSub' enables you to broadcast events from anywhere in your reflex network to
+-- any location that is subscribed to the topic they are published to. Topics are best
+-- representated as constructors of some custom GADT which carries a phantom type to
+-- constrain the messages that can be published to it.
+class (Reflex t, Monad m, Semigroup a) => PubSub t tag m a where
```
endgame, line 28: Looking at the proposals I made below, there's an interesting covariance/contravariance thing going on here. `publishes` is like `contramap` and `subscribes` is kinda like `fmap`. Maybe there's some formulation of this where you can have a `Divisible`/`Decidable`-ish interface to `publish` and an `Applicative`/`Alternative`-ish interface to `subscribe`? (Is it even useful to have this?)

Sketching:
```haskell
-- feels like it could be Contravariant/Divisible/Decidable?
-- conquer = Foo $ const $ pure ()   , perhaps?
-- 'lose' seems to require that you prove the event will never fire?
newtype Foo t m a = Foo { runFoo :: Event t a -> m () }

-- feels like it could be Functor/Applicative/Alternative?
-- Or maybe it's just rehashing the Functor/Apply/Alt instances from semigroupoids? I think you're still unable to write 'pure'.
newtype Bar t m a = Bar { runBar :: m (Event t a) }
```

I don't know if this is interesting enough to bother stalling this PR, or general enough that it should go somewhere else.

src/Reflex/PubSub.hs
JBetz: Yeah, that's a good point. This can effectively recreate the "who wrote to this database table?" problem on your frontend, so you need to be aware of how that effects debuggability. Tracking with types helps some but it still doesn't tell you who published some event without a thorough grep. Also, since the problem it's meant to solve is handling long-range communication, trying to minimize it's scope kind of defeats the purpose, so that doesn't really work as a measure of whether it's the right thing to do.

JBetz:
Thanks for the comments @endgame. I still feel like there's something useful here, if only a symptom of a deeper problem, but after processing your feedback and others I'm not convinced this is the cure.

@ryantrinkle pointed out that you can get the basic idea for this using `QueryT`, where your queries end up looking something like `DMap Topic (Map Query)`. And then I was thinking how the topic could be an index into multiple external datasources a reflex network might be interested in, or slices of a single data source. Or if you have some sort of peer-to-peer setup, you could do `Map Peer (Map Query a)`. 

In any case, there are already abstractions available for the subscription side of things, and since publishing is what's likely to become intractable, I'd rather sort out a better solution for publishing before trying to push forward something as overpowered as pure pubsub. I'm not planning to close this PR quite yet, but probably won't be doing much if any work on it either unless we can fix or at least mitigate the "action at a distance" problem.

endgame:
Another thought: could this be a standalone library instead of being integrated into reflex itself?

PR #434, opened 2020-06-13
Bump version bounds

maralorn:
I have successfully build reflex with those dependencies in nixpkgs.

I would very much appreciate a hackage revision with these version bounds updated.


3noch:
Thank you!

PR #436, opened 2020-06-19
Make build work with newest dependent-map

maralorn:
This haden‘t cropped up before because the version constraint in patch
prevent cabal from ever picking this dependent-map version


3noch:
Excellent. Thank you very much!

PR #437, opened 2020-06-19
Fix hlint 3 test and a GC test

maralorn:
* hlint released a new version
* the Refl constructor does not seem to get exported in newer versions of
  some dependency


ali-abrar:
Thanks! Included in #446

PR #439, opened 2020-06-28
Add mmorph MFunctor instances

kmicklas:
I also cleaned up some `MonadTrans` instances that can be derived, since it's closely related to `MFunctor`.

### DIFF
--- a/src/Reflex/BehaviorWriter/Base.hs
+++ b/src/Reflex/BehaviorWriter/Base.hs
@@ -24,6 +24,7 @@ module Reflex.BehaviorWriter.Base
 import Control.Monad.Exception
 import Control.Monad.Identity
 import Control.Monad.IO.Class
+import Control.Monad.Morph
 import Control.Monad.Reader
 import Control.Monad.Ref
 import Control.Monad.State.Strict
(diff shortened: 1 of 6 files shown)


src/Reflex/BehaviorWriter/Base.hs
```diff
@@ -48,7 +49,18 @@ import Reflex.TriggerEvent.Class
 
 -- | A basic implementation of 'BehaviorWriter'.
 newtype BehaviorWriterT t w m a = BehaviorWriterT { unBehaviorWriterT :: StateT [Behavior t w] m a }
-  deriving (Functor, Applicative, Monad, MonadIO, MonadFix, MonadAsyncException, MonadException) -- The list is kept in reverse order
+  -- The list is kept in reverse order
```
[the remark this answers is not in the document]
3noch: I realize it was already there.

src/Reflex/Query/Base.hs
```diff
@@ -291,8 +295,8 @@ withQueryT f a = do
   return result
 
 -- | Maps a function over a 'QueryT' that can change the underlying monad
-mapQueryT :: (forall b. m b -> n b) -> QueryT t q m a -> QueryT t q n a
-mapQueryT f (QueryT a) = QueryT $ mapStateT (mapEventWriterT (mapReaderT f)) a
+mapQueryT :: Monad m => (forall b. m b -> n b) -> QueryT t q m a -> QueryT t q n a
```
3noch, line 298: This *technically* makes the change breaking...

kmicklas: @3noch Do you think I should put it back? I can also make `hoist = mapQueryT`.

3noch: `hoist = mapQueryT' might be better. The next release will be breaking anyway, but if there's a way to minimize it we might as well. I'm also not actually worried about this particular thing breaking anyone, but I have to err on the side of pedantic.

kmicklas: Done.

3noch:
Thanks @kmicklas!

src/Reflex/BehaviorWriter/Base.hs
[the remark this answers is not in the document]
3noch: This comment has to do with the `StateT [Behavior t w]`

PR #444, opened 2020-10-18
Fixes for ghc 8.10

hamishmack:


ryantrinkle:
Looks like this is just failing for GHC 8.0 and 8.2.  It may be time to drop support for those.

PR #445, opened 2020-10-20
fix: check on headless exit on post build

michivi:
Don't enter the headless main loop if there's a shutdown request upon post build.

Without this check, if the application immediately requests shut down on a post build action, the application fails with a `thread blocked indefinitely in an MVar operation` instead of exiting gracefully.

An example use case is trying to read a file on post build, but failing as the file cannot be found. The application would handle this issue as it sees fit and immediately request exit, but fail to do so.

### DIFF
--- a/reflex.cabal
+++ b/reflex.cabal
@@ -342,6 +342,14 @@ test-suite RequesterT
     Reflex.Plan.Pure
     Test.Run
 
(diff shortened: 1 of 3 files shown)


ryantrinkle:
Beautiful patch; thanks!

Issue #412, opened 2020-04-15
[suggestion] move Test module into main library


pdlla:
The Test.Run module is useful for testing reflex apps as well.

Exposing this test module encourages more modular design where view/controller is cleanly separated from state. For a new reflex developer, finding information on how to test an app w/out rendering it in reflex-dom was a major challenge. Before discovering this, I forked reflex-basic-host and making my own version which is almost identical to whats in Test.Run  https://github.com/pdlla/reflex-basic-host 

On the plus side I now completely understand how reflex hosts are built. But I think this information should be easier for new developers to find. TBH I'd like to see this documented in quickref.

Also thanks for making such a great package!

Ericson2314:
If it brings in too many extra deps we could perhaps make it a "sub library" in the same package.

pdlla:
created a PR: https://github.com/reflex-frp/reflex/pull/450

PR #450, opened 2021-01-16
Add testing host methods

pdlla:
This PR introduces test host module into Reflex to facilitate testing. There are two parts:

- Reflex.Test.SimpleHost.hs - this is basically test/Test/Run.hs converted into a more general use module
- Reflex.Test.Monad.Host.hs - this module contains a monad for firing a reflex network and reading its outputs for the purpose of testing with an imperative syntax

This is a module I've been using already https://github.com/pdlla/reflex-test-host which I think belong in the main reflex library. Although more importantly for myself, I need Reflex.Test.Monad.Host for some of the work I plan on doing for reflex-vty.

I've already been using this library extensively in my project https://github.com/pdlla/potato-flow you can see many examples of usage in the test/*.

There's also a reflex-vty layer built over the testing monad. You can see some basic examples here https://github.com/pdlla/potato-flow-vty (see Reflex/Vty/Test/Monad/Host.hs and test/*)

Note 1: this resolves issue https://github.com/reflex-frp/reflex/issues/412
Note 2: happy to remove hspec dependency

### DIFF
--- a/cabal.project.local
+++ b/cabal.project.local
@@ -1,0 +1,6 @@
+
+package *
+  ghc-location: ghc
+
+program-locations
+  ghc-location: ghc
--- a/reflex.cabal
+++ b/reflex.cabal
@@ -145,6 +145,8 @@ library
     Reflex.Requester.Class,
     Reflex.Spider,
     Reflex.Spider.Internal,
+    Reflex.Test.SimpleHost
+    Reflex.Test.Monad.Host
     Reflex.Time,
     Reflex.TriggerEvent.Base,
     Reflex.TriggerEvent.Class,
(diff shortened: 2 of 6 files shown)


pdlla:
@ali-abrar this PR is for the base test module I've been using. See description for link to reflex-vty test module that's built on top of this.

I'm planning on writing UTs for reflx-vty using this test module so ideally I'd like this to be in the base reflex library so that I don't have to pull in pdlla/reflex-test-host as a dependency into reflex-vty (or some other solution).

Note, admittedly, the test monad is rather convoluted and I have only written very basic tests using it. Still, I think it's the right way to go and about as simple as it can be. I should be writing more complicated test cases soon-ish however I thought it'd be better to open this PR now to get the conversation going.

pdlla:
review pls :)?

pdlla:
making my quarterly check in to see if anyone is available to review :)

Ericson2314:
@pdlla perhaps it is fine, but I would be more confortable of the `Reflex.Test` helpers could go in a private "sub library". Expanding the public interface creates a permanent maintenance, so I am always a bit nervous to do so principle.

pdlla:
@Ericson2314  that's fine with me. And I guess tests can also be moved to separate project so that the main library does not need to depend on it. A review would definitely still be appreciated. 

Somewhat related: Is there any place where reflex libraries are documented? Not just for this library, but for other reflex libraries. That was definitely one of the biggest challenges for me when learning to use reflex--bad discovery on up to date supporting libraries and examples :(.

PR #457, opened 2021-06-16
Merge a few things in from v0.7.1.1

Ericson2314:
- Clean up GitHub action
- Complete change log
- Fix GHCJS tested version
- Bump base lower bound because we in fact do not build with 8.2 and earlier.


tomsmalley (approved):
CI looks like it should have passed

PR #460, opened 2021-11-13
relaxed witherable upper bound to allow 0.4

noinia:
Seems to compile/run fine here on my little reflex-sdl based example


ali-abrar:
Thanks

PR #470, opened 2022-04-28
Fix type-o

Rosuavio:


Ericson2314:
events is plural so I think this is correct?

Rosuavio:
Ah, missed that.

PR #474, opened 2022-06-10
Fix some warnings

Ericson2314:


Ericson2314:
Somebody want to review this? :) I think it is all fine, but it would be good to have a second pair of eyes.

ali-abrar (approved):

PR #484, opened 2022-12-16
Always enable Template Haskell

Ericson2314:
With splices, it should no longer be an issue.


maralorn:
I am just a bit worried because reflex is probably also used by quite a few users who don’t have access to the splice trick (yet)?

Ericson2314:
@maralorn I would imagine people that don't have access to the splicing trick or iserv are probably also not doing the sort of cross compilation where one of those is necessary?

maralorn:
Yeah, probably.

maralorn (approved):

PR #492, opened 2023-04-12
Fix build against ghc 9.6 under '--allow-newer'

alexfmpe:
https://github.com/haskell/mtl/blob/master/CHANGELOG.markdown#23----2022-05-07

https://hackage.haskell.org/package/base-4.18.0.0/docs/Debug-Trace.html#v:traceEventWith
The `since 4.17` annotation suggests we should have hit this with #486 as well but it's [wrong](https://gitlab.haskell.org/ghc/ghc/-/commit/2aa0770845631e4355f55694f49b3e4b66ecf751#cbecf35546dee344ca6fb2e73c7515f07e868d0f_307_306)

### DIFF
--- a/src/Reflex/BehaviorWriter/Base.hs
+++ b/src/Reflex/BehaviorWriter/Base.hs
@@ -21,7 +21,9 @@ module Reflex.BehaviorWriter.Base
   , withBehaviorWriterT
   ) where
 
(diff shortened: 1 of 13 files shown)


ali-abrar:
No longer necessary, given #493. Please reopen if I've missed something.

PR #500, opened 2024-06-10
Reflex.Network: improve haddocks

endgame:
* Replace reference to deprecated function with its replacement
* Normalise comment formatting
* Suggest how to flatten the result of `networkHold`


alexfmpe:
`*promptly` ? That doesn't sound like something we want to recommend outside of very specific situations. I can only remember using it once in 5+ years

endgame:
I haven't used `networkHold` much at all, but the previous version of the docs referenced a `...Promptly` function. I think otherwise you miss any event firings from the frame that swaps the network in, and that's usually something I want (to capture).

ali-abrar (approved):

ali-abrar:
Agree we should rethink the use of promptly here, but that isn't new to this PR.

PR #499, opened 2024-04-11
Build with ghc 9.8

alexfmpe:
Needs a release of https://github.com/reflex-frp/patch/pull/54

PR #502, opened 2024-08-03
Build with ghc(js) 9.10

ymeister:
Needs https://github.com/reflex-frp/patch/pull/56 (or https://github.com/reflex-frp/patch/pull/55).

From `patch` PR:
>When building on ghc js, requires single-core building (https://gitlab.haskell.org/ghc/ghc/-/issues/25083#note_578275).
>
>For nix:
>```
>patch = haskellLib.overrideCabal (drv: { enableParallelBuilding = false; }) super.patch;
>```
>For cabal:
>```
>cabal build -j1
>```


ymeister:
Needs https://github.com/reflex-frp/patch/pull/56 to be in the hackage to pass the CI.

Ericson2314:
Hoping to land https://github.com/reflex-frp/reflex/pull/499 first to separate warnings/cabal cleanup vs new versions a bit.

src/Reflex/Class.hs
```diff
@@ -679,15 +678,12 @@ instance (Reflex t, IsString a) => IsString (Behavior t a) where
 
 instance Reflex t => Monad (Behavior t) where
   a >>= f = pull $ sample a >>= sample . f
-  -- Note: it is tempting to write (_ >> b = b); however, this would result in (fail x >> return y) succeeding (returning y), which violates the law that (a >> b = a >>= \_ -> b), since the implementation of (>>=) above actually will fail.  Since we can't examine 'Behavior's other than by using sample, I don't think it's possible to write (>>) to be more efficient than the (>>=) above.
```
alexfmpe, line 682: @ryantrinkle is this obsolete now that MonadFail got pulled out?

src/Reflex/Optimizer.hs
```diff
@@ -14,26 +14,57 @@ module Reflex.Optimizer
...
+makeInlinable = CoreDoPluginPass "MakeInlinable" $ \modGuts -> do
+  let f v = setIdInfo v $ let i = idInfo v in
+        setInlinePragInfo i $ let p = inlinePragInfo i in
+        if isDefaultInlinePragma p
+        then defaultInlinePragma { inl_inline = Inlinable (inl_src p) }
+        else p
+      newBinds = flip map (mg_binds modGuts) $ \case
+        NonRec b e -> NonRec (f b) e
+        Rec bes -> Rec $ map (first f) bes
+  return $ modGuts { mg_binds = newBinds }
```
alexfmpe, line 64: Why duplicate this?

ymeister: They differ for different GHC versions:

https://github.com/reflex-frp/reflex/blob/0bf60acd31d4eec5b38eb1e9f9fcffcf0598aaab/src/Reflex/Optimizer.hs#L77
vs
https://github.com/reflex-frp/reflex/blob/0bf60acd31d4eec5b38eb1e9f9fcffcf0598aaab/src/Reflex/Optimizer.hs#L47

https://github.com/reflex-frp/reflex/blob/0bf60acd31d4eec5b38eb1e9f9fcffcf0598aaab/src/Reflex/Optimizer.hs#L59
vs
https://github.com/reflex-frp/reflex/blob/0bf60acd31d4eec5b38eb1e9f9fcffcf0598aaab/src/Reflex/Optimizer.hs#L89

You could of course separate the common logic and reuse in both, but IMO time spent on trying to do so is simply not worth it.

src/Reflex/Class.hs
ryantrinkle: Yeah, we could use the simpler implementation now.

src/Reflex/Optimizer.hs
ryantrinkle: We could just delete "Reflex Optimizer" completely.  It never really worked very well.

ali-abrar (approved):

ali-abrar:
I applied the workaround here https://github.com/haskell-actions/setup/issues/77 to the macos jobs

PR #508, opened 2024-09-13
Add note about `requesting` semantics

LightAndLight:
This seems to be the case for Obelisk's HTTP requester. Should it be true in general? Is this the right place to document it?

Being explicit about this matters to me, because it's possible to implement a requester has the request and response events fire at the same time.
If that were the case, then you wouldn't be able to, say, transition a DOM element to "loading" while a request was in flight.


bglgwyng:
#505
Actually, my fix is to handle immediate responses. I'm not certain about the semantics, but it feels buggy when I play with my app. It's very likely that someone will write an immediate response for test purposes.

maralorn:
I have recently had a similar issue. In my opinion users of the class shouldn’t make any assumptions about the timing.

PR #519, opened 2025-01-16
Remove outdated CPP

alexfmpe:


bench/Main.hs
```diff
@@ -29,11 +29,6 @@ main :: IO ()
 main = defaultMain
   [ bgroup "micro" micros ]
 
-#if !(MIN_VERSION_deepseq(1,4,2))
```
alexfmpe, line 32: We require `base >= 4.11` and https://hackage.haskell.org/package/deepseq-1.4.2.0 only supports `< 4.10` so we always use newer versions

src/Data/AppendMap.hs
```diff
@@ -25,11 +25,7 @@ import Prelude hiding (null)
 import Data.Coerce
 import Data.Default
 import Data.Map (Map)
-#if MIN_VERSION_containers(0,5,11)
```
alexfmpe, line 28: We directly require `containers >= 0.6`

src/Reflex/Dynamic/TH.hs
```diff
@@ -90,22 +90,13 @@ mkDynExp s = case Hs.parseExpWithMode Hs.defaultParseMode { Hs.extensions = [ Hs
   Hs.ParseFailed (Hs.SrcLoc _ l c) err -> fail $ "mkDyn:" <> show l <> ":" <> show c <> ": " <> err
   Hs.ParseOk e -> qDynPure $ return $ everywhere (id `extT` reinstateUnqDyn) $ Hs.toExp $ everywhere (id `extT` antiE) e
     where TH.Name (TH.OccName occName) (TH.NameG _ _ (TH.ModName modName)) = 'unqMarker
-#if MIN_VERSION_haskell_src_exts(1,18,0)
```
alexfmpe, line 93: https://hackage.haskell.org/package/haskell-src-exts-1.18.0 has `base < 4.11`

maralorn (approved):
Also great!

PR #523, opened 2025-02-20
headless: Add MonadThrow (Performable m)

ali-abrar:


alexfmpe:
#524 should help with GHA CI

maralorn:
How does this work semantically? What happens if an error is thrown in a performEvent? Will that terminate the reflex network?

PR #524, opened 2025-02-20
Don't fail fast in CI

alexfmpe:


alexfmpe:
We could disable 9.12 on macos until we can get it from ghcup but while we're still a bit flaky in CI I'd rather we always see what ends up passing

ymeister (approved):

PR #528, opened 2026-02-13
Support GHC 9.14

ymeister:


bglgwyng:
I checked that Reflex builds well with GHC 9.14.1 using allow-newer. Do we have any other blockers now?

Issue #490, opened 2023-03-27
Space leak when using the Applicative instance of Behavior


albertov:
I've encountered a space leak in the Reflex library when using the Applicative instance of Behavior. The space leak occurs when a Behavior is constructed with the Applicative instance and sampled, as shown in the following snippet:

```haskell
sample ((<>) <$> current dynA <*> current dynB)
```

However, when the Behavior is constructed by calling current on a Dynamic which is constructed with the Applicative instance, there is no space leak:

```haskell
sample ( current ( (<>) <$> dynA <*> dynB ))
```

I've profiled my program and found that the space leak is related to DEAD_WEAK objects created by the behaviorPull closure and retained by the accumMaybeMDyn closure.

I'm currently working around it by using a forked version of `reflex-vty` which defines `_vtyResult_picture :: Dynamic t V.Picture` instead of `_vtyResult_picture :: Behavior t V.Picture` (see https://github.com/plow-technologies/reflex-vty/commit/e426a01ec52e5c37312ddb4c41c24426d2b4e8f9) but I believe a proper fix belongs in Reflex since the documentation suggests that sampling a `Behavior` for outputs by the host framework is the recommended pattern.

I'm using GHC 9.2.4 and reflex-0.8.2.1 but I've also reproduced it with the `develop` branch of Reflex. I'm attaching the simplest reproducer I could came up with. It can be called with `constant-memory` or `increasing-memory` as an argument, the later demos the space leak. I've also attatched the SVG rendering of  the `.hp ` files for each run (GitHub won't allow the `.hp` files)

![repro-leak-constant-mem](https://user-images.githubusercontent.com/865291/228002833-2104463f-7d7f-4002-87d2-097247c137c6.svg)
![repro-leak-increasing-mem](https://user-images.githubusercontent.com/865291/228002838-a2e6afa0-7b8e-4292-8ab3-e6b73969de4f.svg)

```haskell
{-# LANGUAGE ConstraintKinds #-}
{-# LANGUAGE FlexibleContexts #-}
{-# LANGUAGE LambdaCase #-}
{-# LANGUAGE OverloadedStrings #-}
{-# LANGUAGE RankNTypes #-}
{-# LANGUAGE TypeFamilies #-}

module Main where

import Control.Concurrent
import Control.Monad
import Control.Monad.Fix
import Control.Monad.IO.Class
import Control.Monad.Identity
import Control.Monad.Ref
import Data.Dependent.Sum
import Data.IORef
[...]

PR #533, opened 2026-06-24
Reproduce Applicative Behavior space leak

parenthetical:
Reproduces #490

### DIFF
--- a/reflex.cabal
+++ b/reflex.cabal
@@ -447,6 +447,19 @@ test-suite rootCleanup
       , ref-tf
       , these
 
+test-suite behaviorLeak
+    type:             exitcode-stdio-1.0
+    main-is:          behaviorLeak.hs
+    hs-source-dirs:   test
+    default-language: Haskell2010
+    ghc-options:      -Wall -O2 -rtsopts "-with-rtsopts=-T"
+    build-depends:
+        base >= 4.11 && < 4.22
+      , dependent-sum
+      , reflex
+      , ref-tf
+      , primitive
+
 benchmark spider-bench
     type:             exitcode-stdio-1.0
     main-is:          Main.hs
--- a/test/behaviorLeak.hs
+++ b/test/behaviorLeak.hs
@@ -1,0 +1,67 @@
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeFamilies #-}
+
+-- | Test whether Applicative combination of Behaviors leaks memory when one of
+-- the Behaviors' source events doesn't update.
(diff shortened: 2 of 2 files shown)


ali-abrar:
I also encountered this when working on https://github.com/reflex-frp/reflex-vty/issues/88

In that case there was a more critical space leak in reflex-vty that has been fixed. However, there's a slower space leak that arises out of the issue you've reproduced here. I was seeing a leak that grew with the number of pulls/Behaviors.

When I profiled, I got a lot of DEAD_WEAK entries. I think what was going on there was that the list of invalidators was itself preventing the gc of the dead-weak objects, but I'm not 100% sure yet. My guess here was that if you have a behavior that never changes but is re-sampled every frame because it's part of a combined behavior that *does* get invalidated every frame, that behavior accumulates invalidators without bound. In reflex-vty this happens because you've got a lot of UI elements that aren't changing constantly but are nonetheless behaviors.

ali-abrar:
We could maybe replace this [unconditional consing](https://github.com/reflex-frp/reflex/blob/fa0621c6dad4fe5d35e2ef67b775d637788e3f2f/src/Reflex/Spider/Internal.hs#L712) with something that invalidates and conses, but the performance penalty might be too high: you'd have to check which ones were dead which requires reading the IORef and deferencing the Weak.

ali-abrar:
> We could maybe replace this [unconditional consing](https://github.com/reflex-frp/reflex/blob/fa0621c6dad4fe5d35e2ef67b775d637788e3f2f/src/Reflex/Spider/Internal.hs#L712) with something that invalidates and conses, but the performance penalty might be too high: you'd have to check which ones were dead which requires reading the IORef and deferencing the Weak.

Despite that, perhaps an amortized pruning approach could work here. E.g., we pick some number N that is our pruning threshold. Instead of `IORef [Weak (Invalidator x)]` we keep `IORef (Int, [Weak (Invalidator x)])`. Whenever we cons, we increment the Int. If we're about to cons and the Int exceeds N, we do the per-Weak check and prune operation that we don't want to do every time.

We'd have to pick an N that makes it unlikely that we'll ever do the check and prune on actually active nodes: they're going to be reset frequently enough that we don't ever need to do an element-by-element pruning. The N serves as a cap for static nodes, so that their list of weaks is bounded.

ali-abrar (approved):

parenthetical:
Thanks @ali-abrar. I've benchmarked both the indiscriminate and amortized cleanup and the impact doesn't seem too bad. I don't think the benchmark includes the degenerate cases though, I'll see whether we can add some.

A better solution to this problem might be to use events to keep behaviors up to date, if we can make it work wrt cycles and such.

PR #537, opened 2026-07-03
Use specialized int-based instead of generic merge

parenthetical:
Fix stray mergeG-based implementations of IntMap merging; we have a more efficient mergeInt* family of functions. This should make DynamicWriter slightly more efficient.

### DIFF
--- a/src/Reflex/DynamicWriter/Base.hs
+++ b/src/Reflex/DynamicWriter/Base.hs
@@ -65,9 +65,9 @@ mergeDynIncremental a = unsafeBuildIncremental (mapM (sample . current) =<< samp
   where changedValues = fmap (PatchMap . fmap Just) $ mergeMapIncremental $ mapIncrementalMapValues updated a
         addedAndRemovedValues = flip pushAlways (updatedIncremental a) $ \(PatchMap m) -> PatchMap <$> mapM (mapM (sample . current)) m
 
-mergeIntMapDynIncremental :: Reflex t => Incremental t (PatchIntMap (Dynamic t v)) -> Incremental t (PatchIntMap v)
-mergeIntMapDynIncremental a = unsafeBuildIncremental (mapM (sample . current) =<< sample (currentIncremental a)) $ addedAndRemovedValues <> changedValues
-  where changedValues = fmap (PatchIntMap . fmap Just) $ mergeIntMapIncremental $ mapIncrementalMapValues updated a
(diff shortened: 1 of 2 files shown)


src/Reflex/DynamicWriter/Base.hs
```diff
@@ -65,9 +65,9 @@ mergeDynIncremental a = unsafeBuildIncremental (mapM (sample . current) =<< samp
   where changedValues = fmap (PatchMap . fmap Just) $ mergeMapIncremental $ mapIncrementalMapValues updated a
         addedAndRemovedValues = flip pushAlways (updatedIncremental a) $ \(PatchMap m) -> PatchMap <$> mapM (mapM (sample . current)) m
 
-mergeIntMapDynIncremental :: Reflex t => Incremental t (PatchIntMap (Dynamic t v)) -> Incremental t (PatchIntMap v)
-mergeIntMapDynIncremental a = unsafeBuildIncremental (mapM (sample . current) =<< sample (currentIncremental a)) $ addedAndRemovedValues <> changedValues
-  where changedValues = fmap (PatchIntMap . fmap Just) $ mergeIntMapIncremental $ mapIncrementalMapValues updated a
+mergeIntDynIncremental :: Reflex t => Incremental t (PatchIntMap (Dynamic t v)) -> Incremental t (PatchIntMap v)
+mergeIntDynIncremental a = unsafeBuildIncremental (mapM (sample . current) =<< sample (currentIncremental a)) $ addedAndRemovedValues <> changedValues
+  where changedValues = fmap (PatchIntMap . fmap Just) $ mergeIntIncremental $ mapIncrementalMapValues updated a
```
ymeister, line 70: Why the top-level name change? it still takes and outputs`PatchIntMap`s.

parenthetical: It's not exported; I just changed it because the merge it uses is called `mergeInt`, and the deprecated merge is `mergeIntMap`.

ymeister: Dunno, the signature of the named function still operates on PatchInt\_Map\_

parenthetical: I'm simply making it analogous to the existing functions, this one is a `Reflex` class member:
```
  mergeIntIncremental :: Incremental t (PatchIntMap (Event t a)) -> Event t (IntMap a)
```
As you can see it also has a `PatchIntMap`.

